PR #75, opened 2024-02-04
Rating chart updates & stat fixes

hburn7:
Closes #39 

@AkinariHex There are breaking changes.

`/api/stats/{playerId}` & `/api/stats/{username}`:
- There is a new field `ratingChart` (replaces `ratingStats`). Each item is still grouped into the same "1 collection per day" format.
- The rating chart now shows whether the item is an adjustment. Any item with `isAdjustment = true` means the rating change comes from something other than an osu! match
- Removed `mostPlayedTeammateName`, `mostPlayedOpponentName`, and `bestTeammateName` from `matchStats` field.

```json
  "ratingChart": {
    "chartData": [
      [
        {
          "name": "S1AT: (Peanut Overcharge) vs (average osu enjoyers)",
          "matchId": 37546,
          "matchOsuId": 96891231,
          "matchCost": 1.4739258926651206,
          "ratingBefore": 699.824077356349,
          "ratingAfter": 749.7960149715763,
          "volatilityBefore": 225,
          "volatilityAfter": 224.30826246700434,
          "ratingChange": 49.97193761522726,
          "volatilityChange": -0.6917375329956599,
          "isAdjustment": false
        }
      ],
      [
        {
          "name": "S1AT: (osu!zhdayu) vs (Peanut Overcharge)",
          "matchId": 37566,
          "matchOsuId": 97194288,
          "matchCost": 1.465160329925816,
          "ratingBefore": 749.7960149715763,
          "ratingAfter": 799.114430922645,
          "volatilityBefore": 224.30826246700434,
          "volatilityAfter": 223.57843078967647,
          "ratingChange": 49.3184159510688,
          "volatilityChange": -0.7298316773278657,
          "isAdjustment": false
        }
      ],
```

Here is an example of decay:
```json
[
        {
          "name": "Decay",
          "matchId": null,
          "matchOsuId": null,
          "matchCost": null,
          "ratingBefore": 1014.5539768123309,
          "ratingAfter": 1011.8539768123309,
          "volatilityBefore": 188.5423886396073,
          "volatilityAfter": 188.97151191099871,
          "ratingChange": -2.7000000000000455,
          "volatilityChange": 0.42912327139140416,
          "isAdjustment": true
        }
      ],
      [
        {
          "name": "Decay",
          "matchId": null,
          "matchOsuId": null,
          "matchCost": null,
          "ratingBefore": 1011.8539768123309,
          "ratingAfter": 1009.1539768123308,
          "volatilityBefore": 188.97151191099871,
[...]

PR #106, opened 2024-02-17
Refactor IPlayerRepository method names & docs

hburn7:
Refactors IPlayerRepository to have better method naming, favoring overloads where possible. Also adds documentation to the method calls. IPlayerService has not yet been updated to match the refactored method calls.

Additionally, there is one breaking change:
- Removed MostPlayedTeammateName and MostPlayedOpponentName from AggregatePlayerMatchStatsDTO (part of PlayerStatsDTO).

@AkinariHex please let me know when the above breaking change is safe to merge.


hburn7:
Breaking change is a duplicate of the same change in #75

bugg86 (approved):

hburn7:
I believe the breaking change here has already been addressed.

Issue #85, opened 2024-02-11
Support API versioning


hburn7:
Currently, we have no API versioning. All routes should begin with `/api/v1` instead. This allows us to maintain backwards compatibility with legacy clients, should we choose to update things.

PR #107, opened 2024-02-17
Add versioning to all Controller routes

foxtail286:
Partial fix to Issue #85 (I believe so? I'm new to GH, please correct me)


bugg86:
When you're developing you should be making a branch based on the develop branch and not master. To keep the commit history clean on this PR I would recommend deleting it, and applying the changes to a new branch based on develop.

foxtail286:
Oh, alright. I'll fix that.

PR #108, opened 2024-02-17
Add versioning to all Controller routes

foxtail286:
Partial fix to Issue #85

### DIFF
--- a/API/Controllers/BeatmapsController.cs
+++ b/API/Controllers/BeatmapsController.cs
@@ -8,7 +8,7 @@ namespace API.Controllers;
 
 [ApiController]
 [EnableCors]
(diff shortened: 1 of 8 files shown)


hburn7:
Thanks for looking into this. The pattern I'd rather use is like this:

```cs
[ApiVersion("1")]
[Route("api/{v:apiVersion}/[controller]")]
```

This will require more setup, including adding the `Asp.Versioning.Mvc v8.0.0` package - still need to research myself, but it'll be more work than just a plain text edit.

hburn7:
Going to close this now since this needs a good bit of work and I don't want it to grow stale. Feel free to write here on any progress updates (or if you decide to no longer pursue this). Thanks.

PR #110, opened 2024-02-20
OAuth2 Implementation

### LINKED ISSUE #89: Refactor authentication scheme to use oAuth2-like strategy
We need to refactor the authentication/authorization portion to move away from a single "privileged secret" and instead use a proper strategy.

I believe the right approach is as follows:
- Create a new database table that stores all API clients and who they belong to, including the per-minute request ratelimit and any special privileges.
- Set ratelimits to the service as a whole from the perspective of the client. I'm not sure quite how .NET does this but this is something we can research.

hburn7:
# OAuth2 Refactor

Closes #89 

This PR refactors the login flow to allow for the following:
- Users authorizing with osu! now have access to a more secure authorization flow.
- We now support third-party OAuth2 clients in a secure fashion.

## Breaking Changes
- Deleted `/api/login` controller, replaced with `/api/oauth` controller.
- Authorization header must now be sent as `Authorization bearer <accessToken>`
- `/api/me/validate` returns `204` if the user is whitelisted, returns `403` otherwise.
- User must have the `whitelist` scope in the database (and thus have this encoded in their `accessToken`) for `/api/me/validate` to return a success code.

## Other Changes
- Added [OAuth2](https://datatracker.ietf.org/doc/html/rfc6749) standard for logging in by using access & refresh tokens.
- Replaced HTTP-Only cookie JWT with a standard response.
- Added ability for users to create third-party clients using a `clientId` and `clientSecret` pattern.
- New database table: `oauth_clients`
- Renamed table field: `users.roles` to `users.scopes`.
- The separation of concerns between users and clients is already accounted for, i.e. we can specify on a per-endpoint level whether a user or client should be given access to this endpoint.
- Configuration breaking change: `Jwt:Issuer` is now named `Jwt:Audience`

Requirements before merging:

`otr-web`:
- [x] Implement `/api/oauth/authorize?code=<string>` instead of `/api/login?code=<string>`. The website **must** track the expiration of the `accessToken` correctly. When the `accessToken` expires, the client (`otr-web` in this case) **must** call `/api/oauth/refresh?refreshToken=<refreshToken>` in order to get a new `accessToken`.
- [x] Store the `refreshToken` in an HTTP-Only cookie on the client-side.
- [x] Store the `accessToken` in a standard cookie, as it is not necessary for this to be as secure. Name this cookie `OTR-Access-Token`.
- [x] Pass the `accessToken` as such in the `Authorization` header: `Authorization: Bearer <accessToken>`
- [x] When the `refreshToken` expires, the API will return `401`. At such point, the user must login via osu! again.
This produces a response as follows:
```js
{
    "accessToken": "<long string>",
    "refreshToken": "<long string>",
    "accessExpiration": int
}
```


hburn7:
I may have introduced a bug where the OTR-Access-Token cookie does not properly authenticate a user, if no other authorization header is provided. Will investigate.

hburn7:
> I may have introduced a bug where the OTR-Access-Token cookie does not properly authenticate a user, if no other authorization header is provided. Will investigate.

As a matter of fact, I did this intentionally as I believe it would be best practice for the client to *always* pass the `accessToken` as `Authorization: Bearer <accessToken>` because the server should not be looking for any specific cookies. The *client* can, however, and use that cookie as a means to remember what the authorization header should be.

bugg86 (commented):
While conceptually some of this OAuth stuff still goes over my head, the code itself looks good and I can see how it all connects together. Do you want me to test this locally before approving?

hburn7:
> While conceptually some of this OAuth stuff still goes over my head, the code itself looks good and I can see how it all connects together. Do you want me to test this locally before approving?

I expect Akinari will do most of the testing, but if you'd like to, you can.

bugg86 (approved):

bugg86 (approved):

AkinariHex (dismissed):
It is working, some bugs where addressed and fixed.
Right now not having issues, once you fix the conflics you can merge the PR

PR #122, opened 2024-02-22
Implements Rating Util Tests

### LINKED ISSUE #14: Unit tests for RatingUtils
Currently, RatingUtils is untested. We need to ensure that the ranking class names and rating values are a set amount.

Create separate unit tests for each of the items in RatingUtils. The class can be `RatingUtilsTests`. 

For example,
```cs
[Test]
public void Bronze_Rating_Is499() {
    Assert.That(RatingUtils.RatingBronze, Is.EqualTo(499));
}
```

Do the same for ranking names (bronze/silver/etc.)

o-saund:
Implemented tests for the RatingUtils.cs class - Improving test coverage to 99%

one line is not covered which is an unreachable case discussed with @hburn7 earlier

Let me know if more test cases are needed, or need to be changed, etc. Thanks.

closes #14


APITests/Utilities/RatingUtilsTests.cs
```diff
@@ -110,4 +110,152 @@ public void ProvisionalFlag_IsFalse_WhenNotProvisional(double volatility, int ma
...
+	public void GetRatingDelta_ReturnsRatingDelta_GivenCurrentRating(double rating, double expectedDelta)
+	{
+		// Arrange
+		
+		// Act
+		double delta = RatingUtils.GetRatingDeltaForNextTier(rating);
+		
+		// Assert
+		Assert.Equal(expectedDelta, delta);
+	}
```
hburn7, line 223: Here I would utilize different parameters because these `InlineData`s look a bit confusing. i.e. `(double rating, double offset)`

I'd even recommend using `RatingUtils.GetNextTier` (or some similar method) to get the next tier, then apply the delta in the `// Act` portion.

Also, just declaring variables is something that goes in the `// Arrange` section, `// Act` is for like calculations or other logic. So I would request this be addressed in the rest of your tests.

APITests/Utilities/RatingUtilsTests.cs
```diff
@@ -110,4 +110,152 @@ public void ProvisionalFlag_IsFalse_WhenNotProvisional(double volatility, int ma
...
+	[InlineData(RatingUtils.RatingDiamondII + 1, RatingUtils.RatingDiamondII)]
+	[InlineData(RatingUtils.RatingDiamondI + 1, RatingUtils.RatingDiamondI)]
+	[InlineData(RatingUtils.RatingMasterIII + 1, RatingUtils.RatingMasterIII)]
+	[InlineData(RatingUtils.RatingMasterII + 1, RatingUtils.RatingMasterII)]
+	[InlineData(RatingUtils.RatingMasterI + 1, RatingUtils.RatingMasterI)]
+	[InlineData(RatingUtils.RatingGrandmasterIII + 1, RatingUtils.RatingGrandmasterIII)]
+	[InlineData(RatingUtils.RatingGrandmasterII + 1, RatingUtils.RatingGrandmasterII)]
+	[InlineData(RatingUtils.RatingGrandmasterI + 1, RatingUtils.RatingGrandmasterI)]
+	[InlineData(RatingUtils.RatingEliteGrandmaster + 1, RatingUtils.RatingEliteGrandmaster)]
+	public void GetPreviousTier_ReturnsPreviousTier_GivenCurrentTier(double rating, double? expectedPrevTier)
```
hburn7, line 141: Here you can just pass in the `rating + 1` as inline data (rating), then in the `// Act` portion, specify `rating - 1` as the `expectedPrevTier`. Saves you some headache.

hburn7 (changes requested):
Looks good, we can clean some things up though.

hburn7:
@myssto / @bugg86 Can either of you pick this up?

hburn7 (approved):

PR #127, opened 2024-02-25
feature/edit match validation

### LINKED ISSUE #91: Allow modifying of the match verification status
Introduce `/api/matches/verification-status PATCH` to accept changing of the match's `MatchVerificationStatus` value.

bugg86:
Closes #91

Adds a new PATCH endpoint for editing the match verification status of a match.

## Other Changes
- A few miscellaneous code warning cleanups.
- Added helper methods from MatchService -> MatchRepository to edit just the match verification status value.

### DIFF
--- a/API/Repositories/Interfaces/IMatchesRepository.cs
+++ b/API/Repositories/Interfaces/IMatchesRepository.cs
@@ -15,7 +15,7 @@ public interface IMatchesRepository : IRepository<Match>
 	Task<Match?> GetFirstMatchNeedingAutoCheckAsync();
 	Task<IList<Match>> GetNeedApiProcessingAsync();
 	Task<IEnumerable<Match>> GetByMatchIdsAsync(IEnumerable<long> matchIds);
-	Task<Match> UpdateExisting(MatchDTO matchDto);
+	Task<Match> UpdateVerificationStatus(int id, int? verificationStatus);
 
 	/// <summary>
 	///  Used to queue up matches for verification.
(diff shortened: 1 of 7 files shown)


API/Repositories/Interfaces/IMatchesRepository.cs
```diff
@@ -15,6 +15,7 @@ public interface IMatchesRepository : IRepository<Match>
 	Task<Match?> GetFirstMatchNeedingAutoCheckAsync();
 	Task<IList<Match>> GetNeedApiProcessingAsync();
 	Task<IEnumerable<Match>> GetByMatchIdsAsync(IEnumerable<long> matchIds);
+	Task<Match> UpdateVerificationStatus(int id, int? verificationStatus);
```
hburn7, line 18: `verificationStatus` should be an `int` on update (not `int?`). I'm honestly not sure if there is ever a case where `verificationStatus` is null, will have to investigate separately.

hburn7 (changes requested):

API/Repositories/Interfaces/IMatchesRepository.cs
bugg86: Gotcha, I think changing the method to accept an int and using the null forgiving operator when passing it in should probably be fine for now. Like this: `(int)match.VerificationStatus!`.

hburn7: Null ignore should be fine but doesn't feel optimal.

hburn7: I'll check db and see if this field should be nullable.

hburn7:
Confirmed that `matchVerificationStatus` can be made non-null. There are no null values in the database.

bugg86:
> Confirmed that `matchVerificationStatus` can be made non-null. There are no null values in the database.

Alright do you want me to change the entity to be non-null in this PR?

hburn7 (approved):

hburn7:
> > Confirmed that `matchVerificationStatus` can be made non-null. There are no null values in the database.
> 
> Alright do you want me to change the entity to be non-null in this PR?

Considering everything is working correctly, let's address in a separate PR.

PR #132, opened 2024-02-27
Add configuration validation, use IOptions

### LINKED ISSUE #131: Load Configs at Startup
Configs need to be loaded and validated at startup to reduce issues when building and running the project.

myssto:
Added a `ConfigurationExtensions` class with `BindAndValidate` method to validate configuration values before startup to be used while adding services.

Now register each configuration to the container and validate content on startup.

Replaced each instance of `IConfiguration` with its respective `IOptions<T>` counterpart.

Closes #131


API/Configurations/JwtConfiguration.cs
```diff
@@ -0,0 +1,16 @@
...
+{
+    public class JwtConfiguration
+    {
+        public const string Position = "Jwt";
+
+        [Required(ErrorMessage = "Key is required!")]
+        public string Key { get; set; } = string.Empty;
+
+        [Required(ErrorMessage = "Audience is required!")]
+        [Url]
```
hburn7, line 13: let's put small tags on top

API/Configurations/OsuConfiguration.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System.ComponentModel.DataAnnotations;
+
+namespace API.Configurations
+{
+    public class OsuConfiguration
+    {
+        public const string Position = "Osu";
+
+        [Required(ErrorMessage = "ApiKey is required!")]
+        public string ApiKey { get; set; } = String.Empty;
```
hburn7, line 10: Throughout the codebase we tend to use `string.Empty` instead (adjust where used)

API/Configurations/OsuConfiguration.cs
```diff
@@ -0,0 +1,21 @@
...
+
+        [Required(ErrorMessage = "ClientId is required!")]
+        [Range(1, long.MaxValue, ErrorMessage = "ClientId is required!")]
```
hburn7, line 13: Definitely shouldn't repeat error messages for separate validations.

API/Utilities/ConfigurationExtensions.cs
```diff
@@ -0,0 +1,28 @@
+﻿using System.ComponentModel.DataAnnotations;
+
+namespace API.Utilities
+{
+    public static class ConfigurationExtensions
+    {
+        public static T BindAndValidate<T>(this IConfiguration configuration, string sectionName) where T : new()
```
hburn7, line 7: Assuming the logic is the same, I'd definitely prefer a more concrete reference than `where T : new()`. (something like `where T : class` or some clearer type).

hburn7 (changes requested):

API/Utilities/ConfigurationExtensions.cs
myssto: `where T : class, new()` is also valid, so I went with that

hburn7: Do you have to have the `new()`? I've just never seen it before, feels weird lol.

myssto: I mean I don't *have* to, that's just what was on the SO post so that's what I went with

I'll remove it if you want LOL

hburn7: I see, I guess it's fine. I looked it up, it's because `: class` allows abstract classes I think.

API/Configurations/OsuConfiguration.cs
```diff
@@ -0,0 +1,21 @@
...
+    public class OsuConfiguration
+    {
+        public const string Position = "Osu";
+
+        [Required(ErrorMessage = "ApiKey is required!")]
+        public string ApiKey { get; set; } = string.Empty;
+
+        [Required(ErrorMessage = "ClientId is required!")]
+        [Range(1, long.MaxValue, ErrorMessage = "ClientId must be a positive, non-zero value!")]
+        public long ClientId { get; set; } = 0;
```
hburn7, line 14: I think default of 0 is kinda silly lol, is there any use to setting a default?

API/Configurations/OsuConfiguration.cs
```diff
@@ -0,0 +1,21 @@
...
+
+        [Required(ErrorMessage = "ClientSecret is required!")]
+        public string ClientSecret { get; set; } = string.Empty;
+        public bool AutoUpdateUsers { get; set; } = false;
+        public bool AllowDataFetching { get; set; } = false;
```
hburn7, line 19: bools are false by default right?

hburn7 (changes requested):

API/Configurations/OsuConfiguration.cs
myssto: possible null reference error grrrrrrrrrrrrr

API/Configurations/OsuConfiguration.cs
myssto: Oh I actually didn't know this, you're right they are false by default

API/Configurations/OsuConfiguration.cs
myssto: Ok I went back to look again and I guess it might have been Intellicode bugging because it's not giving me the warning anymore. I'll remove the unnecessary defaults

hburn7 (approved):

bugg86 (approved):
I like this a lot, being able to only give classes access to specific parts of the configuration is really dope.

Issue #113, opened 2024-02-22
Return submitted tournaments with verification statuses


hburn7:
An endpoint needs to be created for returning a user's tournament submissions, along with information about their verification statuses. Here is what needs to be returned:

For every submitted tournament, return:
- The tournament name
- The number of matches submitted by the user for that tournament
- A count of the distinct verification statuses for all matches. (e.g. `countVerified`, `countPreVerified`, and so on for all verification statuses).
- The date of the tournament starting (this is the value of the earliest `match.startTime` field).
- For each match the user submitted under this tournament, return a collection of:
`matchTitle`, `matchVerificationStatus`, `matchStartTime`, `matchUrl` (names are not exact, use the currently existing naming conventions for these fields).

Depends on #112 

@AkinariHex This is intended to serve this section on figma, with adjustments made for tournament collections instead of just listing matches.
![image](https://github.com/osu-tournament-rating/otr-api/assets/38370573/90a2fbcc-f3a1-48b5-a347-484ac1fe4b7f)

PR #139, opened 2024-02-29
Move SubmitterUserId from Match level to Tournament level

### LINKED ISSUE #112: Replace match submission user with tournament submission user
Considering users are only supposed to submit tournaments, not individual matches (and, all individual matches are tied to a parent tournament), we need to identify who the tournament submitter is instead of the match submitter.

This requires modifying the database and a database script to migrate submitter ids. I think it's fine to just take the first match from each tournament, get the `submitted_by_user`, and move the id to the `tournament.submitter_id` field.

`verified_by_user` can stay at the match-level.

myssto:
Closes #112

Move submitter identifiers to `Entities.Tournament` from `Entities.Match`. Uses same naming scheme as previous with Match. `Tournament` now has two new properties, `int? SubmitterUserId` and `User? SubmittedBy`.

Existing values from `match.submitted_by_id` have been populated into `tournaments.submitter_id` in the database.

Tournament objects are now assigned `SubmitterUserId` via `TournamentsRepository.CreateOrUpdateAsync`


API/Entities/Tournament.cs
```diff
@@ -23,6 +23,9 @@ public class Tournament
 	public int Mode { get; set; }
 	[Column("team_size")]
 	public int TeamSize { get; set; }
+	[Column("submitter_id")]
+	public int? SubmitterUserId { get; set; }
+	public User? SubmittedBy { get; set; }
```
hburn7, line 28: I prefer to have the entity references below all the columns in the class file

API/Entities/User.cs
```diff
@@ -25,7 +25,7 @@ public class User
 	[ForeignKey("PlayerId")]
 	[InverseProperty("User")]
 	public virtual Player Player { get; set; } = null!;
-	public IEnumerable<Match>? SubmittedMatches { get; set; }
+	public IEnumerable<Tournament>? SubmittedTournaments { get; set; }
```
hburn7, line 28: Declare `virtual` here.

API/OtrContext.cs
```diff
@@ -240,7 +238,10 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
 			      .HasConstraintName("Tournaments___fkmatchid")
 			      .IsRequired();
 
-			entity.HasIndex(e => new { e.Name, e.Abbreviation }).IsUnique();
+            entity.Property(e => e.SubmitterUserId).IsRequired(false).HasDefaultValue(null);
```
hburn7, line 241: Every tournament has to be submitted by someone.

```suggestion
            entity.Property(e => e.SubmitterUserId).IsRequired();
```

API/OtrContext.cs
```diff
@@ -240,7 +238,10 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
...
+			entity.HasOne(e => e.SubmittedBy).WithMany(u => u.SubmittedTournaments).HasForeignKey(e => e.SubmitterUserId).IsRequired(false);
```
hburn7, line 242:
```suggestion
			entity.HasOne(e => e.SubmittedBy).WithMany(u => u.SubmittedTournaments).HasForeignKey(e => e.SubmitterUserId).IsRequired();
```

hburn7 (changes requested):

myssto:
@hburn7 Considering this hasn't touched prod at all yet, maybe we just scrap the third migration from this and adapt the rest of the code for #142 . The first 2 migrations are perfectly fine and provide enough to work on #113

hburn7:
I've actually got an idea for this.
- Return the match submitter id field as it was
- Keep the tournament submitter id
- Make *both* of these ids immutable
- Have a separate table for match & tournament edits. It could be one table and could be queried by entity type & entity id. It could display which property changed and record the new value. So like this:
```
entity_modifications
- id
- entity_id
- property_name
- value
- updater_id
- timestamp
```

Thoughts?

myssto:
I like that a lot actually! And this way we can keep track of the data history better in case we get a bad batch of changes.

hburn7:
Great, let's do that via this PR then.

hburn7:
Ping me when this is ready for review again.

myssto:
Alright @hburn7 ready to go (it wouldn't let me re-request review for some reason)

API/Repositories/Implementations/MatchHistoryRepository.cs
```diff
@@ -0,0 +1,25 @@
...
+    public MatchHistoryRepository(OtrContext context ,IMapper mapper) : base(context)
+    {
+        _mapper = mapper;
+    }
+
+    public async Task<MatchHistory?> CreateAsync(Match match, HistoryActionType action)
+    {
+        var record = _mapper.Map<MatchHistory>(match);
+        record.HistoryAction = (int)action;
+        // API modifications have an Id of null
```
hburn7, line 21:
```suggestion
        // Modifications made automatically (without user intervention) have an Id of null
```

API/Repositories/Interfaces/IMatchHistoryRepository.cs
```diff
@@ -0,0 +1,15 @@
...
+public interface IMatchHistoryRepository : IRepository<MatchHistory>
+{
+    /// <summary>
+    /// Creates a new <see cref="MatchHistory"/> mapped to the values of <paramref name="match"/> with action type <paramref name="action"/>
+    /// </summary>
+    /// <param name="match"></param>
+    /// <param name="action"></param>
+    /// <returns></returns>
+    Task<MatchHistory?> CreateAsync(Match match, HistoryActionType action);
+}
```
hburn7, line 15: Can you make this generic?

```cs
public interface IHistoryRepository<T> : IRepository<T>
```

Then CreateAsync would take in a T arg, HistoryActionType action.

API/Repositories/Interfaces/IMatchesRepository.cs
```diff
@@ -14,15 +14,9 @@ public interface IMatchesRepository : IRepository<Match>
 	Task<Match?> GetFirstMatchNeedingApiProcessingAsync();
 	Task<Match?> GetFirstMatchNeedingAutoCheckAsync();
 	Task<IList<Match>> GetNeedApiProcessingAsync();
-	Task<IEnumerable<Match>> GetByMatchIdsAsync(IEnumerable<long> matchIds);
+	Task<Match?> GetNoTrackingAsync(int id);
+    Task<IEnumerable<Match>> GetByMatchIdsAsync(IEnumerable<long> matchIds);
```
hburn7, line 18: Rename to `GetAsync` (I've been trying to reduce the amount of overly verbose names in all the repositories).

hburn7 (changes requested):
Some naming changes requested, I also saw a few things that I won't comment on since they'll get picked up by the new formatter.

hburn7:
@myssto These conflicts will be a little tedious to fix.

1. For all files where there are *specifically* only formatting conflicts, you can accept your version. If there are files that have logic changes, you need to merge them manually.
2. When the merges are done, run the formatter.

API/Repositories/Implementations/MatchHistoryRepository.cs
```diff
@@ -0,0 +1,26 @@
...
+    }
+
+    public async Task<MatchHistory?> CreateAsync(Match match, HistoryActionType action)
+    {
+        MatchHistory record = _mapper.Map<MatchHistory>(match);
+        record.HistoryAction = (int)action;
+        // Modifications made automatically (without user intervention) have an Id of null
+        record.ModifierId = null;
+        return await base.CreateAsync(record);
+    }
```
hburn7, line 25: I think this & some other common methods (such as updated [by user]) could go into a HistoryRepositoryBase class and be made generic. Don't want to do this for each history item if the logic is the same.

API/Repositories/Implementations/MatchesRepository.cs
hburn7, line 1: Same here - all of this history stuff can go into a common HistoryRepositoryBase class.

hburn7 (changes requested):

API/Repositories/Implementations/MatchesRepository.cs
hburn7: Yes a history base repo.

API/Repositories/Implementations/HistoryRepositoryBase.cs
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    private async Task<TEntity?> GetNoTrackingAsync(int id) =>
+        await _context.Set<TEntity>().AsNoTracking().FirstOrDefaultAsync(x => x.Id == id);
+
+    private async Task<THistory?> CreateHistoryAsync(int id, HistoryActionType action, int? modifierId = null)
+    {
+        TEntity? origEntity = await GetNoTrackingAsync(id);
+        if (origEntity == null)
+            return null;
```
myssto, line 58: I was thinking this should probably throw an error since realistically if the entity doesn't exist when we get to this point then something probably went wrong, or something is making changes to context outside of the standard CRUD. Which is, and I quote "sussy wussy"

API/Entities/Interfaces/IEntityBase.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace API.Entities.Interfaces;
+
+public interface IEntityBase
```
myssto, line 3: Would this be better as just `IEntity` ?

hburn7: yes

API/Entities/Interfaces/IHistoryEntity.cs
hburn7, line 1: this should inherit `IEntity`

API/Repositories/Implementations/HistoryRepositoryBase.cs
hburn7: I feel like that should be the responsibility of the caller to handle. Returning null *is* the unexpected behavior here.

hburn7 (changes requested):
gonna req changes for now and come back after, haven't reviewed everything.

API/Entities/Interfaces/IUpdateableEntity.cs
hburn7, line 1: This needs to inherit from IEntity and be referenced once in Match.cs (`Match : MatchEntityBase, IUpdateableEntity`)

hburn7 (changes requested):

API/Entities/Interfaces/IUpdateableEntity.cs
myssto: Yea I feel like I knew this... don't know why I did it the way I did lol

hburn7 (approved):

PR #140, opened 2024-03-01
Support Api Versioning

### LINKED ISSUE #85: Support API versioning

bugg86:
Closes #85 

Adds versioning support to every controller. Using Url based version control but we do have the option to use a query or header based version control if we want to later down the road. Deciding exactly how we want to tackle multiple versions of endpoints can come later when we actually want/need to do it. The package is pretty flexible with lots of options so I'm not too concerned about that yet.

This will technically break every endpoint since they'll all need the version added to them, so I'm adding Akinari for review and approval once web side can support it.


hburn7 (dismissed):
Looks solid. Will wait for Akinari.

hburn7 (approved):

Issue #145, opened 2024-03-02
Correct authorization role syntax


hburn7:
Lots of roles are being applied incorrectly. Capitalization is incorrect when compared to the scope documentation. We are also inconsistently enforcing roles.

PR #147, opened 2024-03-02
Update roles across all controllers, remove redundant endpoints

hburn7:
Updates roles (scopes) required across all endpoints to match the [documentation](https://github.com/osu-tournament-rating/otr-wiki/blob/master/api/usage/scopes/en.md)

@AkinariHex Removed a couple of useless endpoints. I don't think any of them are used by you, but just double checking. Additionally, some endpoints may have different permissions that are not expected. Just take a look at all the scopes and let me know if anything looks off.


myssto (commented):
LGTM, will come back and approve once documentation for the wiki is updated [here](https://github.com/osu-tournament-rating/otr-wiki/pull/16)

hburn7:
Closes #145

bugg86 (approved):

myssto (approved):
Because we are making (and planning on making) many breaking changes, in the next week I'm going to say updating the docs isn't as important as merging this for now. I can get around to updating docs when things aren't in flux as much.

PR #155, opened 2024-03-07
Return `T` instead of `T?` in `RepositoryBase.CreateAsync()`

### LINKED ISSUE #109: Do not return `T?` when calling `CreateAsync` on a repository
All database calls to `CreateAsync<T>` return `T?`. Instead, `T` should be returned, as an exception would be thrown by entity framework if something goes wrong during the creation process.

myssto:
`RepositoryBase<T>.CreateAsync(T entity)` now returns `<T>` instead of `<T?>`. All instances where an entity was created and then nullish-ly coalesced have been corrected.

Closes #109


API/Repositories/Implementations/RepositoryBase.cs
```diff
@@ -13,9 +13,11 @@ protected RepositoryBase(OtrContext context)
         _context = context;
     }
 
-    public virtual async Task<T?> CreateAsync(T entity)
+    public virtual async Task<T> CreateAsync(T entity)
     {
-        var created = (await _context.Set<T>().AddAsync(entity)).Entity;
+        T? created = (await _context.Set<T>().AddAsync(entity)).Entity;
+        if (created == null)
+            throw new Exception($"Failed to create {nameof(T)} entity");
```
hburn7, line 20: hmm kinda annoyed the styler is letting this through. I prefer distinct braces for every control flow change.

API/Repositories/Implementations/TournamentsRepository.cs
```diff
@@ -197,13 +197,6 @@ private async Task<Tournament> CreateFromWrapperAsync(TournamentWebSubmissionDTO
...
         };
-
-        var result = await CreateAsync(tournament);
-        if (result == null)
-        {
-            throw new Exception("Tournament could not be created.");
-        }
-
-        return result;
+        return await CreateAsync(tournament);
```
hburn7, line 200:
```suggestion

        return await CreateAsync(tournament);
```

hburn7 (changes requested):

hburn7 (approved):

PR #157, opened 2024-03-07
Fix unapplied styling

hburn7:
Applies all styles via `dotnet format --severity info --verify-no-changes --exclude **/Migrations/* `. Note, there are a good amount of manual fixes that need to be made before merging.


hburn7:
ugh. made this PR from the wrong base.

PR #168, opened 2024-03-08
Add versatile player fetching

### LINKED ISSUE #166: `/players/{id}/info` should scan for multiple matches simultaneously
Replace the following code with a single endpoint.
```cs
[HttpGet("{osuId:long}/info")]
    public async Task<ActionResult<PlayerInfoDTO?>> GetByUserIdAsync(long osuId)
    {
        var info = await _playerService.GetAsync(osuId);

        if (info == null)
        {
            return NotFound($"User with osuid {osuId} does not exist");
        }

        return info;
    }

    [HttpGet("{username}/info")]
    public async Task<ActionResult<PlayerInfoDTO?>> GetByUserIdAsync(string username)
    {
        var player = await _playerService.GetAsync(username);
        if (player != null)
        {
            return Ok(player);
        }

        return NotFound($"User with username {username} does not exist");
    }
```

This endpoint should look for a `key` of type `object`. This key can be anything, i.e. a string, long, or int. We can parse the key in different ways and search the database for player id, player username, and player osu id.

If one match, return it.

If multiple matches, return in the following order: player id match -> osu id match -> username match.

hburn7:
Allows searching for username, osu id, and player id all in the same endpoint.

`/players/{id}/info` & `/players/{username}/info` are replaced by `/players/{key}/info`.

This is not a breaking change and has been tested against otr-web develop.

Closes #166


API/Controllers/PlayersController.cs
```diff
@@ -30,31 +30,19 @@ public async Task<IActionResult> GetAllAsync()
         return Ok(players);
     }
 
-    [HttpGet("{osuId:long}/info")]
-    public async Task<ActionResult<PlayerInfoDTO?>> GetByUserIdAsync(long osuId)
+    [HttpGet("{key}/info")]
+    public async Task<ActionResult<PlayerInfoDTO?>> GetAsync(string key)
```
myssto, line 34:
```suggestion
    [EndpointSummary("Get player info by versatile search")]
    [EnpointDescription("Get player info searching first by id, then osuId, then username")]
    public async Task<ActionResult<PlayerInfoDTO?>> GetAsync(string key)
```

Don't forget your documentation >:)

myssto (changes requested):
LGTM, just add endpoint decorators

API/Controllers/PlayersController.cs
hburn7: so huge

PR #170, opened 2024-03-09
Update formatting settings

myssto:
Use `var` keyword for implied types in assignment. Add husky pre-commit hooks for automated formatting.


hburn7 (approved):
good to go once conflicts are solved

Issue #152, opened 2024-03-06
Remove `/me/validate` endpoint


hburn7:
This is no longer being used.

bugg86:
Looks like this was removed in #183

PR #183, opened 2024-03-11
Fix refreshed access tokens not receiving `user` or `client` claims

### LINKED ISSUE #179: Fix swagger not applying the API version
Swagger does not apply v1 automatically to the endpoints. Swagger should be set to use v1 as the default.

myssto:
Since these claims are manually applied on creation, they need to be manually applied on refresh as well. This pull also addresses two smaller issues, populating the api version in swagger and removing the `/me/validate` endpoint (hence the breaking change tag).

Closes #179, #152

### DIFF
--- a/API/Handlers/Implementations/OAuthHandler.cs
+++ b/API/Handlers/Implementations/OAuthHandler.cs
@@ -172,7 +172,7 @@ public class OAuthHandler : IOAuthHandler
             accessToken = GenerateAccessToken(
                 user.Id.ToString(),
                 _jwtAudience,
-                user.Scopes,
+                [.. user.Scopes, "user"],
                 ACCESS_DURATION_SECONDS
             );
         }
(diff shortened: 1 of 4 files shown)


API/Handlers/Implementations/OAuthHandler.cs
```diff
@@ -172,7 +172,7 @@ IUserRepository userRepository
             accessToken = GenerateAccessToken(
                 user.Id.ToString(),
                 _jwtAudience,
-                user.Scopes,
+                [.. user.Scopes, "user"],
```
hburn7, line 175: Surprised this syntax works, didn't even know this was a thing.

hburn7 (approved):

API/Handlers/Implementations/OAuthHandler.cs
myssto: Yea me neither lol... I found it because the formatter automatically formatted `client.Scopes.Append("client").ToArray();` to `[.. client.Scopes, "client"];` so I guess we have to use it :P

PR #184, opened 2024-03-11
Update all uses of configuration values to instead use `IOptions<T>`

### LINKED ISSUE #182: Use configuration binding for `Auth` section
The `Auth` section of appsettings needs to be bound and validated on load just like the other sections. We need to add this binding to the DI container for the services that need it, namely `OAuthHandler` as it still injects the entire configuration.

Tasks
- [x] Create `/API/Configurations/AuthConfigurations.cs`
- [x] Replace any instances of config values being parsed directly from `IConfiguration configuration`

myssto:
Remove unused config fields `Auth.WebLoginAuthSecret` and `Auth.PrivilegedClientSecret`.
Added a configuration binding for `Auth` section.
Updated all uses of config values that weren't already using `IOptions<T>`

Closes #182

### DIFF
--- a/API/Configurations/AuthConfiguration.cs
+++ b/API/Configurations/AuthConfiguration.cs
@@ -1,0 +1,11 @@
+using System.ComponentModel.DataAnnotations;
+
+namespace API.Configurations;
+
+public class AuthConfiguration
+{
+    public const string Position = "Auth";
+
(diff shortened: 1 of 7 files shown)


hburn7 (approved):
Remind me to change the deploy configs before merging.

myssto:
👍

PR #186, opened 2024-03-12
Add unique instance claim to all tokens.

hburn7:
Adds a unique "instance" claim with a new GUID value to all generated tokens.

* Fixes bug where duplicate tokens could be generated if requested too quickly.


myssto (approved):
This is very smart actually, nice work! I was trying to brainstorm what we could do to add randomness to avoid double generation but this seems perfect. Sidenote we should probably standardize whether claims going forward are to be PascalCase or camelCase.

myssto (approved):

Issue #136, opened 2024-02-29
Use consistent decorators and return values for endpoint methods


myssto:
Every method used as an endpoint within controllers should have a return value matching what the endpoint actually returns, not just `Task<IActionResult>`. These methods should also, at the bare minimum, employ the `[ProducesResponseType]`, `[EndpointSummary]`, and `[EndpointDescription]` decorators for consistency.

hburn7:
After some research, I do agree we should not return the interface and instead return a concrete result. However, instead of ActionResult<T>, a recently introduced design pattern uses TypedResult<T>. I think the latter would be better, seems ActionResult is more on the legacy side nowadays.

hburn7:
I don't know whether using a concrete return type and also using the attribute for ProducesResponseType<T> for 200 is necessary.

myssto:
Yea, probably not necessary to use `ProducesResponseType` when using concrete return types, but a lot of methods are currently missing both

myssto:
Using [this](https://learn.microsoft.com/en-us/aspnet/core/web-api/action-return-types?view=aspnetcore-8.0) document as reference. After some more research and testing, as Stage suggested it looks like the best solution would be to go with using a return type of `Results<T>` (Defined [here](https://learn.microsoft.com/en-us/aspnet/core/web-api/action-return-types?view=aspnetcore-8.0#resultstresult1-tresultn-type)). This approach leverages the static `TypedResults` class, and allows us to forego usage of the `ProducesResponseTypeAttribute`.

Consider the current endpoint `GET` `/api/v1/me`
```cs
    [HttpGet]
    [EndpointSummary("Gets the logged in user's information, if they exist")]
    [ProducesResponseType<UserInfoDTO>(StatusCodes.Status200OK)]
    [ProducesResponseType(StatusCodes.Status401Unauthorized)]
    [ProducesResponseType(StatusCodes.Status404NotFound)]
    public async Task<ActionResult<UserInfoDTO>> GetLoggedInUserAsync()
```

Disregarding the fact that this incorrectly implements `ActionResult<T>` (As you would not need to include a type when using `[ProducesResponseType(StatusCodes.Status200OK)]`), this could be condensed into the following while maintaining swagger definitions:

```cs
    [HttpGet]
    [EndpointSummary("Gets the logged in user's information, if they exist")]
    public async Task<Results<BadRequest<string>, NotFound<string>, Ok<UserInfoDTO>>> GetLoggedInUserAsync()
[...]

PR #189, opened 2024-03-14
Feature/search api

### LINKED ISSUE #111: Search API
Implement search API

Base: `/api/search`

Accept a `string search` as a query parameter and return a `SearchResponseDTO` object, if items are found. Return `null` if no items are found.

Allow searching for:
- Players by username
- Tournaments by name
- Matches by name

Rank results in order:
- Players (if any)
- Tournaments (if any)
- Matches (if any)

For each of the categories, return `IEnumerable<SearchResponseDTO>`.

`SearchResponseDTO` should contain:
- `Text: string`
- `Url: string?`
- `Thumbnail: string?`

Search results should always link to the otr page for the item, if applicable. For now we can link matches to the osu! MP link.

bugg86:
Closes #111 

Adds new controller with a new endpoint that allows for searching of Matches, Tournaments and Players all by name. Request takes in query parameters for either of the 3 and will return a `SearchResponseDTO` with the following formats:

- Thumbnail is a full link to a player's pfp.
- Url is the path to the official otr page.
- Text will be either be: Tournament name, Match ID (osu id), or Player's username.


API/Controllers/SearchController.cs
```diff
@@ -0,0 +1,65 @@
...
+[ApiController]
+[ApiVersion(1)]
+[EnableCors]
+[Authorize(Roles = "system")] // Internal access only at this time
+[AllowAnonymous]
+[Route("api/v{version:apiVersion}/[controller]")]
+public class SearchController(ITournamentsService tournamentsService, IMatchesService matchesService, IPlayerService playerService) : Controller
+{
+    [HttpGet]
+    public async Task<IActionResult> Test()
```
hburn7, line 19: Should be removed

API/Controllers/SearchController.cs
```diff
@@ -0,0 +1,65 @@
...
+    {
+        return Ok();
+    }
+
+    [HttpPost]
+    public async Task<IActionResult> SearchByNames([FromQuery(Name = "tournamentName")] string? tournamentName, [FromQuery(Name = "matchName")] string? matchName, [FromQuery(Name = "username")] string? username)
```
hburn7, line 25: The `Name = ""` is not necessary on `FromQuery` if it's identical to the method param.

API/Controllers/SearchController.cs
```diff
@@ -0,0 +1,65 @@
...
+        if (username is null)
+        {
+            return Ok(null);
+        }
+
+        PlayerDTO? playerDto = await playerService.GetByUsernameAsync(username);
+
+        return playerDto is null ? Ok(null) : Ok(new SearchResponseDTO()
+        {
+            Text = playerDto.Username ?? "Restricted",
```
hburn7, line 59: This matches the behavior of leaderboard results when the username is missing.
```suggestion
            Text = playerDto.Username ?? "<Unknown>",
```

API/Repositories/Implementations/MatchesRepository.cs
```diff
@@ -38,6 +38,13 @@ public async Task<Match> UpdateVerificationStatus(int id, int? verificationStatu
         return existing;
     }
 
+    public async Task<Match?> GetByNameAsync(string name)
```
hburn7, line 41: Because this is a search method, we want to allow partial records. This should be renamed to `SearchAsync(string search)` and query for something like `WHERE match.Name LIKE search (ignore case)`. If there's anything else we want to search for (such as by match id), we should handle this here.

API/Repositories/Implementations/PlayerRepository.cs
```diff
@@ -169,7 +169,7 @@ private IQueryable<Player> Search(string username)
             );
         }
 
-        return _context.Players.Where(p => p.Username != null && p.Username.Equals(username, StringComparison.CurrentCultureIgnoreCase));
+        return _context.Players.Where(p => p.Username != null && p.Username.ToLower().Equals(username.ToLower()));
```
hburn7, line 172: Should handle partial searching.

API/Repositories/Implementations/TournamentsRepository.cs
```diff
@@ -15,7 +15,7 @@ public class TournamentsRepository(OtrContext context) : RepositoryBase<Tourname
     private readonly OtrContext _context = context;
 
     public async Task<Tournament?> GetAsync(string name) =>
-        await _context.Tournaments.FirstOrDefaultAsync(x => x.Name.Equals(name, StringComparison.CurrentCultureIgnoreCase));
+        await _context.Tournaments.FirstOrDefaultAsync(x => x.Name.ToLower().Equals(name.ToLower()));
```
hburn7, line 18: Should handle partial searching.

hburn7 (changes requested):
A good amount of changes need to be made on search logic, will finish review after these changes are made.

API/Controllers/SearchController.cs
```diff
@@ -0,0 +1,65 @@
...
+    [HttpGet]
+    public async Task<IActionResult> Test()
+    {
+        return Ok();
+    }
+
+    [HttpPost]
+    public async Task<IActionResult> SearchByNames([FromQuery(Name = "tournamentName")] string? tournamentName, [FromQuery(Name = "matchName")] string? matchName, [FromQuery(Name = "username")] string? username)
+    {
+        if (tournamentName is not null)
```
myssto, line 27: Prefer `string.IsNullOrEmpty()`

API/Repositories/Implementations/PlayerRepository.cs
myssto, line 172: Refer to my comment in dev chat the other day. LINQ will automatically determine how to compare strings based on the collation of the column that property is mapped to. Prefer usage of just `string.Equals()` and don't worry about messing with case.

myssto (changes requested):
Stage already good points on some of the logic, but these consistencies need to be addressed

API/Controllers/SearchController.cs
bugg86: Fixed in latest commit

API/Controllers/SearchController.cs
```diff
@@ -0,0 +1,58 @@
...
+using Microsoft.AspNetCore.Cors;
+using Microsoft.AspNetCore.Mvc;
+
+namespace API.Controllers;
+
+[ApiController]
+[ApiVersion(1)]
+[EnableCors]
+[Authorize(Roles = "system")] // Internal access only at this time
+[AllowAnonymous]
```
myssto, line 14: Make sure this gets removed. I am pretty sure this bypasses the above `[Authorize(Roles = "system")]`

API/Controllers/SearchController.cs
```diff
@@ -0,0 +1,58 @@
...
+[Route("api/v{version:apiVersion}/[controller]")]
+public class SearchController(ITournamentsService tournamentsService, IMatchesService matchesService, IPlayerService playerService) : Controller
+{
+    [HttpPost]
```
myssto, line 18:
```suggestion
    [HttpGet]
```
This should be a `GET` method

myssto, line 18: Make sure to document this with `EndpointSummaryAttribute` and `EndpointDescriptionAttribute` for clarity. I usually use EndpointSummary for a very brief 1 sentence description, and include an EndpointDescription for more in-depth info.

API/Repositories/Implementations/MatchesRepository.cs
myssto, line 41: Stage's comment is still relevant. I think `SearchAsync` is more proper since the result wouldn't fit as an overload to `GetAsync`. Adjust this where present.

.husky/pre-commit
myssto, line 1: Not sure why this would have been populated blank for you. Next time you merge develop into this branch make sure you accept the incoming change

myssto (changes requested):
Logic looks ok to me now. @hburn7 I was thinking that the code for validating the (tournament, match, user) DTO and creating the new `SearchResponseDTO` looks weird to add into the controller itself. It should probably be moved into it's own `SearchService`?

bugg86:
> Logic looks ok to me now. @hburn7 I was thinking that the code for validating the (tournament, match, user) DTO and creating the new `SearchResponseDTO` looks weird to add into the controller itself. It should probably be moved into it's own `SearchService`?

I was honestly thinking about doing this originally, but it didn't strike me as something that was complex enough to justify creating an entire new service. If we potentially (we probably will) expand this controller, it would probably end up making more sense to create a SearchService.

API/Controllers/SearchController.cs
bugg86: Oops, I did this for other endpoints I made just completely forgot to add them here

hburn7:
Yes definitely create a new service. Logic like that should not be in the controller, even if it's one method.

API/Controllers/SearchController.cs
```diff
@@ -0,0 +1,58 @@
...
+    public async Task<IActionResult> SearchByNames([FromQuery] string? tournamentName, [FromQuery] string? matchName, [FromQuery] string? username)
```
hburn7, line 19: Method return type should be concrete (e.g. `ActionResult<T>`)

API/Controllers/SearchController.cs
```diff
@@ -0,0 +1,58 @@
...
+
+        PlayerDTO? playerDto = await playerService.GetByUsernameAsync(username);
+
+        return playerDto is null ? Ok(null) : Ok(new SearchResponseDTO()
+        {
+            Text = playerDto.Username ?? "<Unknown>",
+            Url = $"/users/{playerDto.Id}",
+            Thumbnail = $"a.ppy.sh/{playerDto.OsuId}"
+        });
+    }
```
hburn7, line 57: Extract to service class.

API/Repositories/Implementations/MatchesRepository.cs
```diff
@@ -38,6 +38,13 @@ public async Task<Match> UpdateVerificationStatus(int id, int? verificationStatu
...
+    {
+        Match? match = await MatchBaseQuery(true).FirstOrDefaultAsync(x => EF.Functions.Like(x.Name, name));
+
```
hburn7, line 44:
```suggestion
```

API/Repositories/Implementations/PlayerRepository.cs
```diff
@@ -163,13 +163,13 @@ private IQueryable<Player> Search(string username)
             return _context.Players.Where(p =>
                 p.Username != null
                 && (
-                    p.Username.Equals(username, StringComparison.CurrentCultureIgnoreCase)
-                    || p.Username.Equals(username, StringComparison.CurrentCultureIgnoreCase)
+                    EF.Functions.Like(p.Username, username)
+                    || EF.Functions.Like(p.Username, username)
```
hburn7, line 167: [See here ](https://stackoverflow.com/a/45709434), I think you need a wildcard (`%`) directly after the search input, and you probably have to do .ToLower(), not sure.

If you find this is not the case in local testing, lmk.

(This applies everywhere `EF.Functions.Like` is used)

API/Repositories/Interfaces/IMatchesRepository.cs
```diff
@@ -11,6 +11,7 @@ public interface IMatchesRepository : IHistoryRepository<Match, MatchHistory>
     Task<IEnumerable<Match>> GetAsync(IEnumerable<long> matchIds);
     Task<IEnumerable<int>> GetAllAsync(bool filterInvalidMatches);
     Task<Match?> GetByMatchIdAsync(long matchId);
+    Task<Match?> GetByNameAsync(string name);
```
hburn7, line 14: See comment on `SearchAsync` method naming

API/Services/Implementations/MatchesService.cs
```diff
@@ -175,6 +175,13 @@ public async Task<IEnumerable<int>> GetAllIdsAsync(bool onlyIncludeFiltered)
         return _mapper.Map<MatchDTO?>(match);
     }
 
+    public async Task<MatchDTO?> GetByNameAsync(string name)
+    {
+        Match? match = await _matchesRepository.GetByNameAsync(name);
+
```
hburn7, line 181:
```suggestion
```
These new lines are cringe!

API/Services/Implementations/PlayerService.cs
```diff
@@ -86,4 +87,11 @@ public class PlayerService(IPlayerRepository playerRepository, IMapper mapper) :
 
     public async Task<PlayerInfoDTO?> GetAsync(string username) =>
         _mapper.Map<PlayerInfoDTO?>(await _playerRepository.GetAsync(username));
+
+    public async Task<PlayerDTO?> GetByUsernameAsync(string username)
+    {
+        Player? player = await _playerRepository.GetAsync(username);
+
```
hburn7, line 94:
```suggestion
```

API/Services/Implementations/TournamentsService.cs
```diff
@@ -30,6 +31,13 @@ public async Task<IEnumerable<TournamentDTO>> GetAllAsync()
         return _mapper.Map<IEnumerable<TournamentDTO>>(items);
     }
 
+    public async Task<TournamentDTO?> GetByName(string name)
+    {
+        Tournament? tournament = await _repository.GetAsync(name);
+
```
hburn7, line 37:
```suggestion
```

API/Services/Interfaces/IMatchesService.cs
```diff
@@ -13,6 +13,7 @@ public interface IMatchesService
 
     Task<IEnumerable<int>> GetAllIdsAsync(bool onlyIncludeFiltered);
     Task<MatchDTO?> GetByOsuIdAsync(long osuMatchId);
+    Task<MatchDTO?> GetByNameAsync(string name);
```
hburn7, line 16: SearchAsync

API/Services/Interfaces/IPlayerService.cs
```diff
@@ -12,6 +12,7 @@ public interface IPlayerService
     Task<int?> GetIdAsync(long osuId);
     Task<int?> GetIdAsync(int userId);
     Task<long?> GetOsuIdAsync(int id);
+    Task<PlayerDTO?> GetByUsernameAsync(string username);
```
hburn7, line 15: Get implies equality - this is a search right?

API/Services/Interfaces/ITournamentsService.cs
```diff
@@ -17,6 +17,7 @@ public interface ITournamentsService
 
     public Task<bool> ExistsAsync(string name, int mode);
     Task<IEnumerable<TournamentDTO>> GetAllAsync();
+    Task<TournamentDTO?> GetByName(string name);
```
hburn7, line 20: Get implies equality - this is a search right?

hburn7 (changes requested):

API/Controllers/SearchController.cs
myssto: Good catch, came back to this just specifically to check but you already got it hehe

bugg86: I think we decided to return ActionResults and then specify the return type in a ProducesResponseType attribute, I added the attribute in a commit I'll be pushing soon.

API/Repositories/Implementations/PlayerRepository.cs
bugg86: Technically case sensitivity on lookups on a column should be handled by collations at the database level, we can use .ToLower() but it could have performance impacts later on from what I understand.

API/Controllers/SearchController.cs
hburn7: @myssto Honestly can't remember if we defined this concretely. What do you think about returning `Task<ActionResult<T>>` *and* documenting the `[ProducesResponseType]` attribute? Can you look and see if swagger can detect just from the ActionResult<T>?

myssto: Will look into it 👍

bugg86: image

Looks like swagger will pickup the return object from the `[ProducesResponseType]` attribute

myssto: I actually found a decent solution to this last night. Take a look at #136 and #207 for a more in depth explanation. The short of it is that we should be using a return type of the new `TypedResponse<T>`

bugg86: Alright will refactor this endpoint to comply with the new standard we're using

bugg86: Updated return types

API/Repositories/Implementations/PlayerRepository.cs
```diff
@@ -166,12 +167,13 @@ private IQueryable<Player> Search(string username)
             return _context.Players.Where(p =>
                 p.Username != null
                 && (
-                    p.Username.ToLower() == username.ToLower()
-                    || p.Username.ToLower() == username.ToLower())
+                    EF.Functions.Like(p.Username, $"%{username}%")
+                    || EF.Functions.Like(p.Username, $"%{username}%")
```
hburn7, line 171: Should be `ILike`

API/Repositories/Implementations/TournamentsRepository.cs
```diff
@@ -167,7 +170,7 @@ LIMIT @count
 
     private async Task<Tournament> UpdateExisting(TournamentWebSubmissionDTO wrapper)
     {
-        Tournament? existing = await GetAsync(wrapper.TournamentName) ?? throw new Exception("Tournament does not exist, this method assumes the tournament exists.");
+        Tournament? existing = await SearchSingleAsync(wrapper.TournamentName) ?? throw new Exception("Tournament does not exist, this method assumes the tournament exists.");
```
hburn7, line 173: You should have another repository function that checks the `ILike` *without* a wildcard (`%`) on each end to indicate case-insensitive title matching. Wildcards don't work here because it leaves room for false matches.

Modifying `SearchSingleAsync` to accept a `bool partialMatch` or similar should be sufficient.

hburn7 (changes requested):
For all `SearchAsync` functions, I think there should be a bool parameter for partial matching. If false, the query should match exactly, just case insensitive. If true, it would function like a non-strict search.

API/Controllers/SearchController.cs
```diff
@@ -0,0 +1,27 @@
...
+public class SearchController(ISearchService service) : Controller
+{
+    private readonly ISearchService _searchService = service;
+
+    [HttpGet]
+    [EndpointSummary("Allows for partial or full searching on the names of tournaments, matches and usernames.")]
+    public async Task<Results<NotFound<string>, Ok<List<SearchResponseDTO>>>> SearchByNames([FromQuery] string? tournamentName, [FromQuery] string? matchName, [FromQuery] string? username)
+    {
+        List<SearchResponseDTO>? response = await _searchService.SearchByNameAsync(tournamentName, matchName, username);
+        return response is null ? TypedResults.NotFound("Requested search returned nothing.") : TypedResults.Ok(response);
```
myssto, line 25: For `NotFound` responses that don't contain any relevant information, prefer just `TypedResults.NotFound()` with the standard message.

myssto (changes requested):

API/Repositories/Implementations/TournamentsRepository.cs
bugg86: I get what you mean but still want to clarify, `ILike` is what enables the case insensitivity vs `Like`, and the `%` wrapper enables the partial matching.

bugg86: Wait nvm that's exactly what you said, I just woke up and haven't had enough coffee 🤦

hburn7 (changes requested):
Will look at the rest after these are addressed.

API/Repositories/Implementations/PlayerRepository.cs
```diff
@@ -150,28 +162,46 @@ await _context
...
         if (username.Contains(' '))
         {
             // Look for users with either ' ' or '_' in the name - osu only uses one (i.e. "Red Pixel" cannot coexist with "Red_Pixel")
             return _context.Players.Where(p =>
                 p.Username != null
                 && (
-                    p.Username.ToLower() == username.ToLower()
-                    || p.Username.ToLower() == username.ToLower())
+                    EF.Functions.ILike(p.Username, username)
+                    || EF.Functions.ILike(p.Username, username)
```
hburn7, line 199: still confused, this is identical?

API/Repositories/Implementations/PlayerRepository.cs
```diff
@@ -60,7 +60,24 @@ public async Task<IEnumerable<Player>> GetAsync(bool eagerLoad)
...
+            throw new Exception("More than one player was found.");
+        }
+
+        if (players.Count == 0)
+        {
+            throw new Exception("No players were found.");
+        }
+
+        return players.First();
+    }
```
hburn7, line 80: On second thought, I think this should return a nullable Player.

If the players.Count > 1, you can still throw the exception, but if count is 0, return null. Otherwise, return players.First() still. This is because throwing exceptions for not found is an awful practice.

API/Repositories/Implementations/PlayerRepository.cs
```diff
@@ -150,28 +162,45 @@ await _context
...
+            if (username.Contains(' '))
+            {
+                // Look for users with either ' ' or '_' in the name - osu only uses one (i.e. "Red Pixel" cannot coexist with "Red_Pixel")
+                return _context.Players.Where(p =>
+                    p.Username != null
+                    && (
+                        EF.Functions.ILike(p.Username, $"%{username}%")
+                    )
+                );
+            }
```
hburn7, line 187: ngl, just scrap all of this. If a user is *searching* for someone, they need to handle the spaces and underscores themselves. Even the osu website cannot decipher `[ MILK_Jiang]` from `[_MILK_Jiang]`. Same for line 192, just scrap the if statement and return the ILike.

This code also doesn't do what I just described, it's a duplicate of what's below.

API/Repositories/Implementations/PlayerRepository.cs
```diff
@@ -150,28 +162,45 @@ await _context
...
             // Look for users with either ' ' or '_' in the name - osu only uses one (i.e. "Red Pixel" cannot coexist with "Red_Pixel")
             return _context.Players.Where(p =>
                 p.Username != null
                 && (
-                    p.Username.ToLower() == username.ToLower()
-                    || p.Username.ToLower() == username.ToLower())
+                    EF.Functions.ILike(p.Username, username)
+                )
             );
         }
```
hburn7, line 201: Scrap this

API/Services/Implementations/SearchService.cs
```diff
@@ -0,0 +1,66 @@
...
+        {
+            return null;
+        }
+
+        var players = (await _playerRepository.SearchAsync(username)).ToList();
+
+        if (players.Count == 0)
+        {
+            return null;
+        }
```
hburn7, line 60: I see 4 distinct functions that could be made out of this, one for each if statement. Please implement for quality.

API/Services/Implementations/SearchService.cs
```diff
@@ -0,0 +1,66 @@
...
+
+        returnList.AddRange(players.Select(player => new SearchResponseDTO() { Text = player.Username ?? "<Unknown>", Url = $"/users/{player.Id}", Thumbnail = $"a.ppy.sh/{player.OsuId}" }));
```
hburn7, line 62: This `/users/{player.Id}` part is an impossible bug waiting to happen. It works for now, but really gotta keep this in mind for when we refactor this URL. Maybe even make an issue now so we know it's there.

API/Services/Interfaces/ISearchService.cs
```diff
@@ -0,0 +1,8 @@
+using API.DTOs;
+
+namespace API.Services.Interfaces;
+
+public interface ISearchService
+{
+    Task<List<SearchResponseDTO>?> SearchByNameAsync(string? tournamentName, string? matchName, string? username);
```
hburn7, line 7: Docs needed

hburn7 (changes requested):
Alright, looked through it all. No new surprises from me after this round of comments.

API/Services/Implementations/SearchService.cs
bugg86: https://github.com/osu-tournament-rating/otr-api/issues/218

API/Services/Implementations/SearchService.cs
```diff
@@ -0,0 +1,101 @@
...
+
+        returnList = await SearchPlayersByNameAsync(username);
+
+        if (returnList is not null)
+        {
+            return returnList.ToList();
+        }
+
+        return null;
+    }
```
hburn7, line 37: Our list needs to contain all search results, not just the first type of result. Instead of returning early, we should combine all results. So, instead of returning the `returnList` at each if statement, instead extend that list with the results of the queries. So by the end, `returnList` contains results from all searches.

This function can be made non-nullable, just return the empty list of results if nothing's found.

hburn7 (changes requested):

API/Services/Implementations/SearchService.cs
bugg86: Do we want to allow for multiple types of results within a single return? Because each search parameter is for a specific table, I assumed we would only be searching for a single one per request. If that's not the case I think we should probably do either the first option below or both of them:

- Add a new field to `SearchResponseDTO` like `Type` or some similar name
- Potentially merge all 3 name fields into a single field, and search all 3 tables for that name

bugg86: I think how we make the change you suggested and I listed above would depend on where this endpoint is going to be implemented, so I'm curious about exactly where and how we were planning on utilizing it in the website

hburn7: Yeah do both, this is intended to function like a site-wide search. The user clicks the search bar, looking for 'anything' basically. So if they type "Stage", the player stage could appear, a beatmap with [Extra Stage] could appear, etc.

I think for an item-specific search (such as tournament search by itself) we should create separate endpoints for that. Ideally the codebase is structured in a way where it's not complicated to do both combined and distinct type searches.

bugg86: We commented at the exact same time but you answered my question in your comment :ehehe:. That makes sense, I'll implement both of those things into the logic

hburn7 (approved):

Issue #128, opened 2024-02-25
Define API usage limits


hburn7:
Identify...

* How to ratelimit individual clients
* What the ratelimit should be by default
* Add a `ratelimit_override` `int` field for api clients on the database side.
* Define ratelimit frequency period (i.e. `x` calls per `minute`, `10 minutes` ?)

PR #198, opened 2024-03-15
Introduce rate limits

myssto:
Introduce rate limits to the API

Uses a [Fixed Window](https://learn.microsoft.com/en-us/aspnet/core/performance/rate-limit?view=aspnetcore-8.0#fixed-window-limiter) rate limit with a default of 30 requests per minute. The configuration uses a partitioned scheme, where each User / OAuthClient has their own limit, and all anonymous requests share a limit. Each partition is identified by the `Issuer` field (we assign from `User.Id` or `OAuthClient.Id`) of our JWT access tokens.

The configuration supports encoding rate limit override values into the JWT access tokens. The overrides enable us to grant a modified token limit and refresh window on a per User / OAuthClient basis. The overrides are stored in the `users` and `oauth_clients` tables as JSON representations of the new `RateLimitOverrides` entity to the `rate_limit_overrides` column. To conserve space in the JWT, they are serialized by excluding null values.

If you are interested in reading about how this is all configured, some good articles are here ([1](https://learn.microsoft.com/en-us/aspnet/core/performance/rate-limit?view=aspnetcore-8.0), [2](https://blog.maartenballiauw.be/post/2022/09/26/aspnet-core-rate-limiting-middleware.html), [3](https://devblogs.microsoft.com/dotnet/announcing-rate-limiting-for-dotnet/))


myssto:
Closes #128

hburn7:
> Each partition is identified by the Issuer field (we assign from User.Id or OAuthClient.Id) of our JWT access tokens.

Does your implementation handle the case where the user id and client ids are the same?

API/Configurations/RateLimitConfiguration.cs
```diff
@@ -0,0 +1,19 @@
...
+    /// <summary>
+    /// The total amount of tokens available per <see cref="Window"/>
+    /// </summary>
+    [Range(1, int.MaxValue, ErrorMessage = "PermitLimit must be an integer greater than 1!")]
+    public int PermitLimit { get; } = 30;
+    /// <summary>
+    /// The amount of time (in seconds) before the <see cref="PermitLimit"/> is refreshed
+    /// </summary>
+    [Range(1, int.MaxValue, ErrorMessage = "Window must be an integer greater than 1!")]
+    public int Window { get; } = 30;
```
hburn7, line 18: This should be 60 for 30req/min right?

API/Entities/User.cs
```diff
@@ -28,6 +28,8 @@ public class User
     [Column("updated", TypeName = "timestamp with time zone")]
     public DateTime? Updated { get; set; }
 
+    public RateLimitOverrides? RateLimitOverrides { get; set; }
```
hburn7, line 31: Document that the column name is defined in `OtrContext`

API/Entities/OAuthClient.cs
```diff
@@ -20,6 +20,8 @@ public class OAuthClient
     [Column("user_id")]
     public int UserId { get; set; }
 
+    public RateLimitOverrides? RateLimitOverrides { get; set; }
```
hburn7, line 23: Document that the column name is defined in `OtrContext`

myssto:
That is a very valid question that I had made a reminder to check on, but never actually did lol. I have a fix in mind, Ill implement real quick

API/Entities/RateLimitOverrides.cs
```diff
@@ -0,0 +1,13 @@
...
+public class RateLimitOverrides
+{
+    public RateLimitOverrides() { }
+    public RateLimitOverrides(int? permitLimit, int? window)
+    {
+        PermitLimit = permitLimit;
+        Window = window;
+    }
+    public int? PermitLimit { get; set; }
+    public int? Window { get; set; }
```
hburn7, line 12: What is the reason for using two constructors? Column names need to be defined on the properties, the migration uses pascal case for this table.

API/Program.cs
```diff
@@ -71,6 +80,82 @@
...
+builder
+    .Services.AddRateLimiter(options =>
+    {
+        RateLimitConfiguration rateLimitConfiguration = builder.Configuration.BindAndValidate<RateLimitConfiguration>(
+            RateLimitConfiguration.Position);
+
+        var defaultFixedOptions = new FixedWindowRateLimiterOptions()
+        {
+            PermitLimit = 30,
+            Window = TimeSpan.FromSeconds(60),
```
hburn7, line 92: I originally was confused by this var name, definitely rename to something like `anonymousRateLimiterOptions`. I thought you were choosing not to use the `rateLimitConfiguration` for some reason.

API/Program.cs
```diff
@@ -71,6 +80,82 @@
...
+
+        options.GlobalLimiter = PartitionedRateLimiter.Create<HttpContext, string>(context =>
+        {
+            // Try to get access token
+            var accessToken =
+                context.Features.Get<IAuthenticateResultFeature>()?.AuthenticateResult?.Properties
+                    ?.GetTokenValue("access_token")?.ToString() ?? string.Empty;
+
+            // Shared partition for anonymous users
+            if (accessToken.IsNullOrEmpty())
```
hburn7, line 125: use `string.IsNullOrEmpty` here.

API/Program.cs
```diff
@@ -71,6 +80,82 @@
...
+            {
+                return RateLimitPartition.GetFixedWindowLimiter(
+                    "anonymous",
+                    _ => defaultFixedOptions
+                );
+            }
+
+            // Try parse rate limit override claim
```
hburn7, line 133:
```suggestion
            // Try to parse the 'ratelimitoverrides' claim
```

hburn7: Define a constant for the claim name ('ratelimitoverrides') in Utilities somewhere.

API/Program.cs
```diff
@@ -71,6 +80,82 @@
...
+            if (overrides is null)
+            {
+                return RateLimitPartition.GetFixedWindowLimiter(jwtToken.Issuer, _ => defaultFixedOptions);
+            }
+            return RateLimitPartition.GetFixedWindowLimiter(
+                jwtToken.Issuer,
+                _ => new FixedWindowRateLimiterOptions()
+                {
+                    PermitLimit = overrides.PermitLimit ?? defaultFixedOptions.PermitLimit,
+                    Window = TimeSpan.FromSeconds(overrides.Window ?? rateLimitConfiguration.Window),
```
hburn7, line 152: On one line `defaultFixedOptions` is used, the next line `rateLimitConfiguration` is used...?

API/Utilities/BitLengthAttribute.cs
hburn7, line 1: Needs unit test

hburn7 (changes requested):

API/Configurations/RateLimitConfiguration.cs
myssto: Yep, this got changed when I went from a bucket to fixed window limit, and this was the one spot this didn't get changed. If you look at the diff of the first two commits this was literally just renamed from `TokensPerPeriod` to `Window`. Silly mistake

API/Entities/User.cs
myssto: I considered leaving the column notation but thought that wouldn't make sense considering the mapping ignores it anyway. Will leave a comment

API/Entities/RateLimitOverrides.cs
myssto: So I was modeling my use of JSON columns off of the Microsoft [documentation](https://learn.microsoft.com/en-us/ef/core/what-is-new/ef-core-7.0/whatsnew#json-columns) and npgSQL [documentation](https://www.npgsql.org/efcore/mapping/json.html?tabs=data-annotations%2Cpoco). In my head some of the wording alluded to needing a parameterized constructor in order for EF to handle automatic (de)serialization, but I could be wrong. I included a parameter-less constructor for ease of partial updates. I was not sure if former was required to exist, but I should have tested instead of leaving it in.

API/Program.cs
myssto: Where would you include this? In something like a `Constants.cs` ?

API/Entities/RateLimitOverrides.cs
hburn7: Sure, lmk what you find. Preferably we don't ever want to use constructors on entities. And test the column naming as well (you can verify with `dotnet ef migrations script` before applying to the db).

API/Program.cs
myssto: My thought is that it would make more sense to use the earlier defined values from `defaultFixedOptions` for consistency, but since `defaultFixedOptions.Window` is already parsed into a `TimeSpan` in order to use a nullish coalescence on `overrides.Window` I would need to use the config value instead. For readability I will just use the values from `rateLimitConfiguration` instead.

API/Entities/RateLimitOverrides.cs
myssto: I was going to raise the point that technically this doesn't conform to our definition of what an entity is and I was unsure where to put it. Maybe `/Entities/JsonElements` ? Since it is a nullable object field, the thought is that it would be easily assignable to a user via something like
```cs
if (user.RateLimitOverrides is null)
{
    user.RateLimitOverrides = new() { TokenLimit = 60 }
}
```
However the other side of that coin is that we could just make the `User.RateLimitOverrides` field non-null and just default to having null fields instead, but I thought that saving space in the db would be more important

API/Program.cs
myssto: Nvm, it defo makes more sense to use an extension to the `ClaimTypes` class

myssto: Actually, `System.Security.Claims.ClaimTypes` is a static class, so I went for a new `OtrClaimTypes` instead.

API/Entities/User.cs
```diff
@@ -28,6 +28,12 @@ public class User
     [Column("updated", TypeName = "timestamp with time zone")]
     public DateTime? Updated { get; set; }
 
+    // // Column name and value initialization is handled via OtrContext
```
hburn7, line 31: double comment xd

hburn7 (approved):

myssto:
Make sure to update project status and close #128 after merging :)

hburn7:
Configs updated.

Issue #207, opened 2024-03-17
Refactor controllers and access layers to comply with new standards


myssto:
## Objective
We have introduced some new standards in terms of design and code quality, and I feel they should be consolidated into a single issue to give a more clear goal to work towards. Each endpoint controller and it's relevant services / resource repositories in the API should be refactored to conform to the following:

### Resource oriented design (Discussed in #138)
Our API _is_ a resource API. We are accessing users, tournaments, etc. Our choices of not only routing but method naming should reflect this. [This article](https://cloud.google.com/apis/design/resources) is a great start and should serve as a general guideline.

### Use consistent decorators and return values (Discussed in #223)
It is important for development and code readability for us to clearly define the interactions that occur on each endpoint. We currently use a very inconsistent mix of `IActionResult`, `ActionResult<T>`, and `ProducesResponseTypeAttribute`, which should be refactored into a consistent use of `IActionResult`.

### Prefer overloads in method naming (Discussed in #95, #154)
This one is pretty self explanatory. We want to prefer overloading existing methods over using overly descriptive names. This *also* applies to controller method names and enforcing resource oriented design. Existing methods like `MeController.GetLoggedInUserAsync` should instead be named simply `MeContoller.GetAsync` (See [Standard Methods](https://cloud.google.com/apis/design/standard_methods)).

Tasks:
- [x] Refactor `BeatmapsController`
- [ ] Refactor `LeaderboardsController`
- [ ] Refactor `MatchesController`
- [x] Refactor `MeController`
- [x] Refactor `OAuthController`
- [ ] Refactor `PlayersController`
- [ ] Refactor `StatsController`
- [x] Refactor `TournamentsController`

myssto:
Updated to include new return type spec

myssto:
This is done at this point :D

PR #210, opened 2024-03-17
BeatmapsController refactor

myssto:
Refactor of the beatmaps controller and related services pursuant to #207 

Breaking changes:
- Remove `GET` `/api/v1/beatmaps/all`
- Remove `GET` `/api/v1/beatmaps/{id}
- Add `GET` `/api/v1/beatmaps`
  - [List](https://cloud.google.com/apis/design/standard_methods#list) endpoint
- Add `GET` `/api/v1/beamtaps/{key}`
  - Versatile search endpoint that returns one beatmap searching first by id (primary key), then by osu! beatmap id

### DIFF
--- a/API/Controllers/BeatmapsController.cs
+++ b/API/Controllers/BeatmapsController.cs
@@ -1,9 +1,9 @@
 using API.DTOs;
-using API.Entities;
 using API.Services.Interfaces;
 using Asp.Versioning;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Cors;
+using Microsoft.AspNetCore.Http.HttpResults;
 using Microsoft.AspNetCore.Mvc;
 
 namespace API.Controllers;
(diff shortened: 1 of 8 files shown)


API/Controllers/BeatmapsController.cs
```diff
@@ -1,35 +1,38 @@
...
+    [EndpointSummary("List beatmaps")]
+    public async Task<Ok<IEnumerable<BeatmapDTO>>> ListAsync() =>
+        TypedResults.Ok(await _beatmapService.ListAsync());
 
-    [HttpGet("{beatmapId:long}")]
-    public async Task<ActionResult<Beatmap>> GetByOsuBeatmapIdAsync(long beatmapId)
+    [HttpGet("{key}")]
+    [EndpointSummary("Get a beatmap by versatile search")]
+    [EndpointDescription("Get a beatmap searching first by id, then by osu! beatmap id")]
+    public async Task<Results<NotFound, Ok<BeatmapDTO>>> GetAsync(string key)
```
hburn7, line 28: I feel like if the search is only being done by these two ids, we can use `int key` instead. Saves a parse call, and the database calls become simpler. e.g. if it's an int, send it to the database using `GetAsync`. If that fails, try to fetch by the long variant for osu id.

hburn7 (changes requested):

hburn7 (approved):

PR #213, opened 2024-03-18
MeController Refactor

myssto:
Refactor of the me controller and related services pursuant to #207

Breaking changes:
- UserInfoDTO changes
  - Remove `UserId` field
  - Change `Id` field to use o!TR id instead of osu! id
  - Rename `Roles` field to `Scopes`
  - Add `PlayerId` field for `User.PlayerId`


API/DTOs/UserInfoDTO.cs
```diff
@@ -3,13 +3,16 @@ namespace API.DTOs;
 public class UserInfoDTO
 {
     /// <summary>
-    ///  The playerId of the user
+    /// Id (primary key) of the associated user
     /// </summary>
-    public int? Id { get; set; }
-    public int? UserId { get; set; }
+    public int Id { get; set; }
```
hburn7, line 8: This class does need a `PlayerId` field as well, that's what `Id` was before.

API/Repositories/Implementations/UserRepository.cs
```diff
@@ -7,46 +7,28 @@ namespace API.Repositories.Implementations;
...
 {
-    private readonly ILogger<UserRepository> _logger = logger;
     private readonly OtrContext _context = context;
 
     public override async Task<User?> GetAsync(int id) =>
         await UserBaseQuery().FirstOrDefaultAsync(u => u.Id == id);
 
-    public async Task<User?> GetForPlayerAsync(int playerId) =>
-        await UserBaseQuery().FirstOrDefaultAsync(u => u.PlayerId == playerId);
+    public async Task<User?> GetByOsuIdAsync(int osuId) =>
```
hburn7, line 17: `osuId` should always be passed as `long`, make sure this is being done everywhere.

API/Services/Implementations/UserService.cs
```diff
@@ -20,19 +20,18 @@ public class UserService(IUserRepository repository) : IUserService
...
-            Username = user.Player.Username,
-            Roles = user.Scopes
+            OsuUsername = user.Player.Username
         };
     }
 
     public async Task<UserInfoDTO?> GetForPlayerAsync(int playerId)
     {
-        User? user = await _repository.GetForPlayerAsync(playerId);
+        User? user = await _repository.GetByOsuIdAsync(playerId);
```
hburn7, line 35: This doesn't make sense. The `playerId` is not the same as `osuId`.

hburn7 (changes requested):
There's a lot of confusion with how the different ids work together. Will re-review after these comments are addressed.

myssto:
Yep I definitely managed to confuse myself along the way here, will give it another go through

myssto:
Should have taken care of the type miss-matching stuff. If you could scrutinize this one a little more than usual for me because I can't seem to find any refactoring errors anymore, but I just want to be 100%

hburn7 (approved):

hburn7:
Awaiting approval before merge:

- [ ] https://github.com/osu-tournament-rating/otr-web/pull/144

PR #214, opened 2024-03-19
Add ability to modify client ratelimits

hburn7:
* Adds ability to modify client ratelimits via `/api/v1/clients/ratelimit` `PATCH`.

### DIFF
--- a/API/Controllers/ClientsController.cs
+++ b/API/Controllers/ClientsController.cs
@@ -19,13 +19,13 @@ public class ClientsController(IOAuthClientService clientService) : Controller
     [HttpPatch("{id:int}/ratelimit")]
     [Authorize(Roles = "admin")]
     [EndpointSummary("Patches the ratelimit for a given client")]
-    public async Task<Results<BadRequest, NotFound<string>, Ok<OAuthClientDTO>>> PatchRatelimitAsync(int id, [FromBody] JsonPatchDocument<RateLimitOverrides> patchedOverrides)
+    public async Task<Results<BadRequest, NotFound, Ok<OAuthClientDTO>>> PatchRatelimitAsync(int id, [FromBody] JsonPatchDocument<RateLimitOverrides> patchedOverrides)
     {
         OAuthClientDTO? currentClient = await clientService.GetAsync(id);
 
         if (currentClient == null)
         {
-            return TypedResults.NotFound("Client not found");
+            return TypedResults.NotFound();
         }
 
         RateLimitOverrides overrides = currentClient.RateLimitOverrides;
--- a/API/Controllers/ClientsController.cs
+++ b/API/Controllers/ClientsController.cs
@@ -1,0 +1,42 @@
+using API.DTOs;
+using API.Entities;
+using API.Services.Interfaces;
+using Asp.Versioning;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Cors;
+using Microsoft.AspNetCore.Http.HttpResults;
+using Microsoft.AspNetCore.JsonPatch;
+using Microsoft.AspNetCore.Mvc;
+
+namespace API.Controllers;
+
+[ApiController]
+[ApiVersion(1)]
+[EnableCors]
+[Route("api/v{version:apiVersion}/[controller]")]
+public class ClientsController(IOAuthClientService clientService) : Controller
+{
+    [HttpPatch("{id:int}/ratelimit")]
+    [Authorize(Roles = "admin")]
+    [EndpointSummary("Patches the ratelimit for a given client")]
+    public async Task<Results<BadRequest, NotFound, Ok<OAuthClientDTO>>> PatchRatelimitAsync(int id, [FromBody] JsonPatchDocument<RateLimitOverrides> patchedOverrides)
+    {
+        var overrides = new RateLimitOverrides();
+
+        if (patchedOverrides.Operations.Any(op => op.op != "replace"))
+        {
+            return TypedResults.BadRequest();
+        }
+
+        patchedOverrides.ApplyTo(overrides, ModelState);
+
+        OAuthClientDTO? client = await clientService.SetRatelimitOverridesAsync(id, overrides);
+
+        if (client == null)
+        {
+            return TypedResults.NotFound();
+        }
+
+        return TypedResults.Ok(client);
+    }
+}
--- a/API/Controllers/ClientsController.cs
+++ b/API/Controllers/ClientsController.cs
@@ -19,9 +19,16 @@ public class ClientsController(IOAuthClientService clientService) : Controller
     [HttpPatch("{id:int}/ratelimit")]
     [Authorize(Roles = "admin")]
     [EndpointSummary("Patches the ratelimit for a given client")]
-    public async Task<Results<BadRequest, NotFound, Ok<OAuthClientDTO>>> PatchRatelimitAsync(int id, [FromBody] JsonPatchDocument<RateLimitOverrides> patchedOverrides)
+    public async Task<Results<BadRequest, NotFound<string>, Ok<OAuthClientDTO>>> PatchRatelimitAsync(int id, [FromBody] JsonPatchDocument<RateLimitOverrides> patchedOverrides)
     {
-        var overrides = new RateLimitOverrides();
+        OAuthClientDTO? currentClient = await clientService.GetAsync(id);
+
+        if (currentClient == null)
+        {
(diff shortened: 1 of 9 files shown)


API/Repositories/Implementations/OAuthClientRepository.cs
```diff
@@ -24,4 +20,20 @@ public async Task<bool> ValidateAsync(int clientId, string clientSecret)
 
         return match != null;
     }
+
+    public async Task<OAuthClient?> SetRatelimitOverridesAsync(int clientId, RateLimitOverrides rateLimitOverrides)
+    {
+        OAuthClient? match = await _context.OAuthClients.FirstOrDefaultAsync(x => x.Id == clientId);
```
myssto, line 26:
```suggestion
        OAuthClient? client = await GetAsync(clientId);
```

Use built in `GetAsync`

API/Repositories/Implementations/OAuthClientRepository.cs
```diff
@@ -10,11 +10,7 @@ namespace API.Repositories.Implementations;
 public class OAuthClientRepository(OtrContext context) : RepositoryBase<OAuthClient>(context), IOAuthClientRepository
 {
     private readonly OtrContext _context = context;
-
-    public async Task<bool> SecretInUseAsync(string clientSecret)
-    {
-        return await _context.OAuthClients.AnyAsync(x => x.Secret == clientSecret);
-    }
+    public async Task<bool> SecretInUseAsync(string clientSecret) { return await _context.OAuthClients.AnyAsync(x => x.Secret == clientSecret); }
```
myssto, line 13: Prefer overload of `ExistsAsync(string clientSecret)`.

API/Controllers/OAuthController.cs
```diff
@@ -74,4 +76,20 @@ public class OAuthController(IOAuthHandler oAuthHandler) : Controller
 
         return TypedResults.Ok(result);
     }
+
+    [HttpPost("ratelimit")]
+    [Authorize(Roles = "admin")]
+    [EndpointDescription("Sets new ratelimit overrides for a given client")]
+    public async Task<Results<BadRequest, Ok<OAuthClientDTO>>> SetRatelimitOverridesAsync([FromQuery] int clientId, [FromQuery] int window, [FromQuery] int limit)
```
myssto, line 83: This should be exported to either `PATCH` `/users/*/clients/*` or `PATCH` `/clients/*`.

API/Controllers/OAuthController.cs
```diff
@@ -74,4 +76,20 @@ public class OAuthController(IOAuthHandler oAuthHandler) : Controller
...
+    {
+        OAuthClientDTO? result = await clientService.SetRatelimitOverridesAsync(clientId,
+            new RateLimitOverrides { PermitLimit = limit, Window = window });
+
+        if (result == null)
+        {
+            return TypedResults.BadRequest();
```
myssto, line 90:
```suggestion
            return TypedResults.NotFound();
```
Make sure return type matches

API/Repositories/Implementations/OAuthClientRepository.cs
```diff
@@ -24,4 +20,20 @@ public async Task<bool> ValidateAsync(int clientId, string clientSecret)
 
         return match != null;
```
myssto, line 21:
```suggestion
        return client != null;
```
Refactor instances where `client` is referred to as `match`

myssto (changes requested):
This endpoint should be a `PATCH` endpoint. I think OAuthController should be kept specific to authorization methods / client creation and not resource access / manipulation methods.

API/Repositories/Implementations/OAuthClientRepository.cs
hburn7: I think I botched this, but I think this works - there is a `SecretInUseAsync` method for the secret. Will implement `ExistsAsync(int clientId, string clientSecret)`.

API/Controllers/ClientsController.cs
```diff
@@ -0,0 +1,42 @@
...
+[EnableCors]
+[Route("api/v{version:apiVersion}/[controller]")]
+public class ClientsController(IOAuthClientService clientService) : Controller
+{
+    [HttpPatch("{id:int}/ratelimit")]
+    [Authorize(Roles = "admin")]
+    [EndpointSummary("Patches the ratelimit for a given client")]
+    public async Task<Results<BadRequest, NotFound, Ok<OAuthClientDTO>>> PatchRatelimitAsync(int id, [FromBody] JsonPatchDocument<RateLimitOverrides> patchedOverrides)
+    {
+        var overrides = new RateLimitOverrides();
```
myssto, line 24: We should be pulling the client's existing overrides here instead of creating a new object. As written this gives no option for a partial update.

myssto (changes requested):

API/DTOs/OAuthClientDTO.cs
```diff
@@ -1,8 +1,11 @@
+using API.Entities;
+
 namespace API.DTOs;
 
 public class OAuthClientDTO
 {
     public int ClientId { get; set; }
     public string ClientSecret { get; set; } = null!;
     public string[] Scopes { get; set; } = [];
+    public RateLimitOverrides RateLimitOverrides { get; set; } = new();
```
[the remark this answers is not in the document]
hburn7: I think it's fine, the alternative is us manually telling them their new ratelimit (which can change at anytime based on what we have in the db, etc.). @bugg86 thoughts?

Also, clients being able to programmatically tell what their ratelimit is seems helpful to me.

API/Controllers/ClientsController.cs
hburn7: Good point.

API/DTOs/OAuthClientDTO.cs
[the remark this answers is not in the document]
myssto: Didn't think about that but it makes sense. From a user perspective it would be nice.

API/Controllers/ClientsController.cs
```diff
@@ -0,0 +1,44 @@
...
+    [HttpPatch("{id:int}/ratelimit")]
+    [Authorize(Roles = "admin")]
+    [EndpointSummary("Patches the ratelimit for a given client")]
+    public async Task<Results<BadRequest, NotFound<string>, Ok<OAuthClientDTO>>> PatchRatelimitAsync(int id, [FromBody] JsonPatchDocument<RateLimitOverrides> patchedOverrides)
+    {
+        OAuthClientDTO? currentClient = await clientService.GetAsync(id);
+
+        if (currentClient == null)
+        {
+            return TypedResults.NotFound("Client not found");
```
myssto, line 28:
```suggestion
            return TypedResults.NotFound();
```

"Client not found" is inferable by just the error itself. Like I mentioned to Convex, I think we should reserve using custom error messages only for when we need to communicate relevant information that cannot be inferred. Make sure return type matches.

API/Controllers/ClientsController.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        if (patchedOverrides.Operations.Any(op => op.op != "replace"))
+        {
+            return TypedResults.BadRequest();
+        }
+
+        patchedOverrides.ApplyTo(overrides, ModelState);
+
+        // We know client is not null here due to the previous null check
+        OAuthClientDTO? client = await clientService.SetRatelimitOverridesAsync(id, overrides);
```
myssto, line 41: We can just make the return type of `SetRatelimitOverridesAsync` non-nullable. We do this elsewhere in the codebase where a function will throw an error instead of returning null. At this point we are already assuming the client exists so it would make sense.

myssto (changes requested):

API/Controllers/ClientsController.cs
hburn7: Ah right, yep sounds good.

API/Controllers/ClientsController.cs
hburn7: I generally disagree with throwing exceptions for null items. The caller (controller in this case) should be able to gracefully handle these null results, rather than try-catching everything. We shouldn't modify service behavior assuming it will only be called in this fashion by the controller. It just so happens that the `SetRatelimitOverridesAsync` function also makes the exact same null check we make in the controller.

myssto: Gotcha, that's fine with me then

myssto (approved):
If you tested and it's working properly then everything LGTM

PR #222, opened 2024-03-26
Add `/tournaments/{id}` and `/tournaments/{id}/matches` endpoints

hburn7:
Was needing a way to get a set of tournament data for testing the processor, so figured I'd go ahead and implement these. Docs will be implemented at a later point when we are 100% certain on the direction of API docs.


myssto (approved):
Approving this so we can get this functionality available to the processor. Expect much of this to be overwritten soon.

hburn7:
Made redundant by #223

PR #223, opened 2024-03-28
TournamentsController Refactor

myssto:
This one is a massive piece of code quality improvement that I would like to serve as a baseline for what we should be doing going forward.

## Breaking Changes
- **Remove** `POST` `/matches/batch`
- **Remove** `PUT` `/tournaments`
- **Add** `POST` `/tournaments`
  - Will also create matches if ids are provided
- **Add** `GET` `/tournaments/{id}`
- **Add** `PATCH` `/tournaments/{id}`
- **Add** `GET` `/tournaments/{id}/matches`
- **Add** `POST` `/tournaments/{id}/matches`
  - Creates new matches for an existing tournament

The flow for submitting tournaments and matches is now completely coupled to the tournaments controller. The rationale for this decision being that matches cannot exist without being owned by a tournament.

*Note the method used for creating matches on the service and repository level. Since matches are relationally tied to a tournament, you can create new matches at the same time as a new tournament by adding the new match entities to the new tournament entity before sending it to `repostitory.CreateAsync()`. This also works for adding new matches to an existing entity before sending to `repostitory.UpdateAsync()`. The only flaw with this is that it will not use any override methods (i.e. `MatchesRepository.CreateAsync()`), thus we need to make sure we use this with due diligence.

## XML Comments
The project will now generate an XML docfile on build (This generation enables compiler warnings for XML formatting, hence the warnings being disabled *for now*). Swagger will use this XML file to add descriptions to any routes, responses, and models that are publicly exposed.

For controller methods, the method itself should return `IActionResult`, and should be annotated with a `ProducesResponseTypeAttribute` for each possible response *in order of possibility*. Each method needs to be documented with an XML comment. The comment should at the bare minimum include `<summary>` and one `<response code=""> for each possible response. For extra information (like notes on behavior or examples, see `TournamentsController.CreateAsync`), use a `<remarks>` tag.

For any object returned by the API (i.e. DTOs), the object definition should include an XML comment for the class, and each property. These comments are picked up by Swagger and are great for documentation.

Going forward I believe we should make a requirement that *all* class and interface methods be documented with the appropriate XML comments, as I have made sure to do in these changes.

## Data Annotations
You will notice the use of `System.ComponentModel.DataAnnotations` for the modified DTO objects. These data annotations serve two purposes. Most importantly, they allow the Controller to automatically validate the fields of any object we wish to receive from a request. For example when receiving a `TournamentWebSubmissionDTO`, instead of validating input manually, the `Controller.ModelState` will do this for us. If the `Controller.ModelState.IsValid` property is false, we can immediately reject the request without any unnecessary processing. The second purpose for the annotations is for hinting in Swagger. Swagger will recognize the annotations and include them as hints in the schema.

## Use of `201 Created`
This response is not likely to be used anywhere else in the API outside of the TournamentsController. Normally, a post method would return the created resource in its entirety, but the usage of `201 Created` is preferred over `200 OK` when the created resource is not complete at the time of creation. This is applicable to matches in particular because almost all of the data is not complete until the API processes new entities. 

I have implemented a base class and interface for these response DTOs along with a convenient way to dynamically generate resource location information with AutoMapper.

## Notes
If there is anything specific in this PR that I haven't explained thoroughly enough, please do let me know so I can provide an explanation. I hope to have some of the content of this PR message added to a Code Quality page for the otr-wiki/api eventually.

Completes task `Refactor TournamentsController` of #207


hburn7:
> Going forward I believe we should make a requirement that all class and interface methods be documented with the appropriate XML comments, as I have made sure to do in these changes.

For classes that inherit an interface, I believe we should enforce interface documentation, as the class kind of gets hidden by the interface.

hburn7:
> This response is not likely to be used anywhere else in the API outside of the TournamentsController
Anytime we create a new 'thing' no?  Such as a new admin comment on a match.

API/Controllers/TournamentsController.cs
```diff
@@ -1,34 +1,223 @@
...
+    /// </summary>
+    /// <remarks>
+    /// The user submits a tournament (containing metadata and links) to the front-end. They are looking to add new data
+    /// to our database that will eventually count towards ratings.
+    ///
+    /// This post endpoint takes these links, validates them (i.e. checks for duplicates,
+    /// whether the match titles align with osu! tournament naming conventions,
+    /// amount of matches being submitted, etc.).
+    ///
+    /// Assuming we have a good batch, we will mark all of the new items as "PENDING".
```
hburn7, line 48: We mark them as pending unless `verify` is `true`. We'd then mark them as `VERIFIED` immediately.

API/Controllers/TournamentsController.cs
```diff
@@ -1,34 +1,223 @@
...
+    /// The API.Osu.Multiplayer.MultiplayerLobbyDataWorker service checks the database for pending links
```
hburn7, line 49: I think a `<see cref="">` tag should be used for this type reference

API/Controllers/TournamentsController.cs
```diff
@@ -1,34 +1,223 @@
...
+    /// <param name="verify">Optionally verify all included matches</param>
+    /// <response code="401">If verify is true and the User does not have match verification privileges</response>
+    /// <response code="400">If a tournament matching the given name and mode already exists, or the given <see cref="wrapper"/> is malformed</response>
+    /// <response code="201">Returns location information for the created tournament</response>
+    [HttpPost]
+    [ProducesResponseType<ModelStateDictionary>(StatusCodes.Status400BadRequest)]
+    [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+    [ProducesResponseType<TournamentCreatedResultDTO>(StatusCodes.Status201Created)]
+    public async Task<IActionResult> CreateAsync(
+        [FromBody] TournamentWebSubmissionDTO wrapper,
```
hburn7, line 62: Rename `wrapper` to `tournamentSubmission`

API/Controllers/TournamentsController.cs
```diff
@@ -1,34 +1,223 @@
...
+    /// This post endpoint takes these links, validates them (i.e. checks for duplicates,
+    /// whether the match titles align with osu! tournament naming conventions,
+    /// amount of matches being submitted, etc.).
+    ///
+    /// Assuming we have a good batch, we will mark all of the new items as "PENDING".
+    /// The API.Osu.Multiplayer.MultiplayerLobbyDataWorker service checks the database for pending links
+    /// periodically and processes them automatically.
+    /// </remarks>
+    /// <param name="wrapper">Tournament submission data</param>
+    /// <param name="verify">Optionally verify all included matches</param>
```
hburn7, line 53:
```suggestion
    /// <param name="verify">Optionally verify all included matches, assuming the user has permission to do so</param>
```

API/Controllers/TournamentsController.cs
```diff
@@ -1,34 +1,223 @@
...
+        }
+
+        // Determine verification source
+        MatchVerificationSource? verificationSource = null;
+        if (verify)
+        {
+            if (!User.CanVerifyMatches())
+            {
+                return Unauthorized();
+            }
```
hburn7, line 78: I think these if statements can be combined. Then `verificationSource` can be created & assigned in one line.

API/Controllers/TournamentsController.cs
```diff
@@ -1,34 +1,223 @@
...
+    /// <response code="401">If verify is true and the User does not have match verification privileges</response>
+    /// <response code="404">If a tournament matching the given id does not exist</response>
+    /// <response code="201">Returns location information of the created matches</response>
+    [HttpPost("{id:int}/matches")]
+    [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+    [ProducesResponseType(StatusCodes.Status404NotFound)]
+    [ProducesResponseType<IEnumerable<MatchCreatedResultDTO>>(StatusCodes.Status201Created)]
+    public async Task<IActionResult> CreateMatchesAsync(
+        int id,
+        [FromBody] MatchesWebSubmissionDTO wrapper,
```
hburn7, line 173: Rename `wrapper` to something more descriptive.

API/Controllers/TournamentsController.cs
```diff
@@ -1,34 +1,223 @@
...
+                (int?)verificationSource
+            );
+
+        if (result is null)
+        {
+            return NotFound();
+        }
+
+        // Use no location header for multiple creation
+        return Created((string?)null, result);
```
hburn7, line 202: What's going on with the `(string?)null`?

API/Controllers/TournamentsController.cs
```diff
@@ -1,34 +1,223 @@
...
+    }
+
+    /// <summary>
+    /// List all matches from a tournament
+    /// </summary>
+    /// <param name="id">Tournament id</param>
+    /// <response code="404">If a tournament matching the given id does not exist</response>
```
hburn7, line 209: Document the 200 code.

API/Controllers/TournamentsController.cs
```diff
@@ -1,34 +1,223 @@
...
+    [HttpGet("{id:int}/matches")]
+    [ProducesResponseType(StatusCodes.Status404NotFound)]
+    [ProducesResponseType<IEnumerable<MatchDTO>>(StatusCodes.Status200OK)]
+    public async Task<IActionResult> ListMatchesAsync(int id)
     {
-        TournamentDTO res = await _service.CreateOrUpdateAsync(tournament, true);
+        TournamentDTO? result = await _tournamentsService.GetAsync(id);
+        if (result is null)
```
hburn7, line 216: I notice we are inconsistently using `is null` and `== null`. I've been using `== null`, just kind of a habit. Do you think `is null` is better going forward, or is there some best practice out there regarding this?

API/DTOs/CreatedResultBaseDTO.cs
```diff
@@ -0,0 +1,17 @@
+namespace API.DTOs;
+
+/// <summary>
+/// Represents a newly created resource
+/// </summary>
+public class CreatedResultBaseDTO
+{
+    /// <summary>
+    /// Id (primary key) of the resource
```
hburn7, line 9: I think primary key is implied here

API/DTOs/MatchesWebSubmissionDTO.cs
```diff
@@ -0,0 +1,21 @@
...
+
+namespace API.DTOs;
+
+/// <summary>
+/// Represents data used to add matches to an existing tournament
+/// </summary>
+public class MatchesWebSubmissionDTO
+{
+    /// <summary>
+    /// The Id (primary key) of the submitting User
```
hburn7, line 11: primary key is implied

API/DTOs/MatchesWebSubmissionDTO.cs
```diff
@@ -0,0 +1,21 @@
...
+    /// </summary>
+    [Range(0, int.MaxValue)]
```
hburn7, line 13: Makes sense to start from 1, but I'm indifferent to this (gee, if only it was a standard practice to use `uint` for this stuff)

API/DTOs/MatchesWebSubmissionDTO.cs
```diff
@@ -0,0 +1,21 @@
...
+    public int SubmitterId { get; set; }
+
+    /// <summary>
+    /// List of match (mp) ids
```
hburn7, line 17:
```suggestion
    /// List of osu! match ids
```

API/Repositories/Interfaces/ITournamentsRepository.cs
```diff
@@ -5,40 +5,59 @@ namespace API.Repositories.Interfaces;
...
-    /// <param name="wrapper">The user input required for this tournament</param>
-    /// <param name="updateExisting">Whether to overwrite values for an existing occurrence of this tournament</param>
-    /// <returns></returns>
-    public Task<Tournament> CreateOrUpdateAsync(
-        TournamentWebSubmissionDTO wrapper,
-        bool updateExisting = false
-    );
+    /// <param name="id">Primary key</param>
+    /// <param name="full">Whether to eagerly load navigational properties</param>
+    Task<Tournament?> GetAsync(int id, bool full = false);
```
hburn7, line 13: Rename `full` to `eagerLoad`

API/Repositories/Interfaces/IRepository.cs
```diff
@@ -6,36 +6,49 @@ public interface IRepository<T>
...
     /// <summary>
-    /// Updates an entity, returning the number of rows affected.
+    /// Updates an entity
     /// </summary>
+    /// <returns>Number of rows affected</returns>
     Task<int> UpdateAsync(T entity);
 
     /// <summary>
-    ///  Deletes an entity from the database by its ID. If successful, returns the primary key of the entity, otherwise null.
+    /// Deletes an entity from the database by its ID
```
hburn7, line 33: I think `ID` should be `id` instead. This is the first time I've seen it in capitals.

API/Repositories/Interfaces/IRepository.cs
```diff
@@ -6,36 +6,49 @@ public interface IRepository<T>
...
     /// <summary>
-    ///  Returns true if an entity with the given ID exists in the database.
+    /// Returns true if an entity with the given ID exists in the database.
     /// </summary>
     Task<bool> ExistsAsync(int id);
 
     /// <summary>
     /// Bulk inserts a collection of entities into the database.
     /// </summary>
-    /// <param name="entities"></param>
```
hburn7, line 36: I feel we can readd this xml doc and say like `The entities to insert`

API/Repositories/Interfaces/ITournamentsRepository.cs
```diff
@@ -5,40 +5,59 @@ namespace API.Repositories.Interfaces;
...
 
+    /// <summary>
+    /// Returns true if an entity with the given name and mode exists
```
hburn7, line 16:
```suggestion
    /// Returns whether an entity with the given name and mode exists
```

API/Repositories/Interfaces/ITournamentsRepository.cs
```diff
@@ -5,40 +5,59 @@ namespace API.Repositories.Interfaces;
...
+
+    /// <summary>
+    /// Search for a tournament by name
+    /// </summary>
+    public Task<IEnumerable<Tournament>> SearchAsync(string name);
+
+    /// <summary>
+    /// Create team size statistics for a player
+    /// </summary>
+    /// <param name="playerId">Id (primary key) of target player</param>
```
hburn7, line 28: primary key is implied

API/Services/Interfaces/IMatchesService.cs
```diff
@@ -1,9 +1,27 @@
...
+    /// <param name="matchIds">List of match ids</param>
+    /// <param name="verify">Submitter is a match verifier</param>
+    /// <param name="verificationSource">Source of verification (int representation of <see cref="MatchVerificationSource"/></param>
+    /// <returns>Location information for the created matches, or null if parent tournament does not exist</returns>
+    Task<IEnumerable<MatchCreatedResultDTO>?> CreateAsync(
+        int tournamentId,
+        int submitterId,
+        IEnumerable<long> matchIds,
+        bool verify,
+        int? verificationSource
```
hburn7, line 22: I wonder if it's possible to pass in the enum here? Doesn't have to be addressed now, but would eventually like to get rid of all the int casting - the db should be able to handle the type directly in the entity just fine.

API/Utilities/ClaimsPrincipalExtensions.cs
```diff
@@ -43,6 +41,34 @@ public static bool IsUser(this ClaimsPrincipal claimsPrincipal)
     public static bool IsMatchVerifier(this ClaimsPrincipal claimsPrincipal) =>
         IsInRole(claimsPrincipal, "verifier");
 
+    /// <summary>
+    /// Denotes the principle as having either the verifier or admin role
+    /// </summary>
+    /// <param name="claimsPrincipal"></param>
+    /// <returns></returns>
+    public static bool CanVerifyMatches(this ClaimsPrincipal claimsPrincipal) =>
+        (claimsPrincipal.IsAdmin() || claimsPrincipal.IsMatchVerifier());
```
hburn7, line 50: Strictly speaking, only a user with the `verifier` scope should have this ability. In the database, we give roles in a verbose manner (i.e. "admin" does not imply "verifier"). Would like to have a better solution in the future, but for now this should be addressed.

API/Utilities/ClaimsPrincipalExtensions.cs
```diff
@@ -43,6 +41,34 @@ public static bool IsUser(this ClaimsPrincipal claimsPrincipal)
...
+    /// <summary>
+    /// Returns the appropriate <see cref="MatchVerificationSource"/> enum for the principle
+    /// </summary>
+    /// <param name="claimsPrincipal"></param>
+    /// <returns></returns>
+    public static MatchVerificationSource? VerificationSource(this ClaimsPrincipal claimsPrincipal)
+    {
+        if (claimsPrincipal.IsAdmin())
+        {
+            return MatchVerificationSource.Admin;
```
hburn7, line 61: General comment, eventually I want to get rid of this distinction. It makes no difference to us.

hburn7 (changes requested):
whew. got through it.

API/Utilities/ClaimsPrincipalExtensions.cs
myssto: Gotcha, this is based off of some function I thought used to be in the claims principle extensions

API/Repositories/Interfaces/ITournamentsRepository.cs
myssto: Did this initially but changed my mind, guess I didn't catch all the instances

API/Controllers/TournamentsController.cs
myssto: This tag was copied directly from the old matches controller function. I don't believe `<see cref="">` work in swagger but I'll double check.

API/Controllers/TournamentsController.cs
myssto: This was done by the formatter I guess? I will late a look as to why

API/Controllers/TournamentsController.cs
myssto: > The equality operator can be overridden. "is" keyword is a reference check. "==" will almost always give you what you want. "is" will always give you what you want.

I know we aren't possible overriding the equality operator anywhere, but using `is null` over `==` is generally considered safer. Plus as a python dev it's kind of a force of habit.

API/Repositories/Interfaces/IRepository.cs
myssto: Feel its pretty redundant. Param named "entities" implies this.

hburn7: We do a lot of redundant documentation anyway :p

API/Controllers/TournamentsController.cs
hburn7: Sure, I will try to adopt this then. #230

API/Repositories/Implementations/TournamentsRepository.cs
```diff
@@ -17,22 +13,11 @@ public class TournamentsRepository(OtrContext context) : RepositoryBase<Tourname
...
-        {
-            throw new Exception("No tournaments were found.");
-        }
+    public async Task<Tournament?> GetAsync(int id, bool eagerLoad = false) =>
+        eagerLoad ? await TournamentsBaseQuery().FirstOrDefaultAsync(x => x.Id == id) : await base.GetAsync(id);
 
-        return tournaments.First();
-    }
+    public async Task<bool> ExistsAsync(string name, int mode) =>
+        await _context.Tournaments.AnyAsync(x => x.Name.ToLower() == name.ToLower() && x.Mode == mode);
```
hburn7, line 20: I feel like we should use `ILike` here

hburn7 (approved):
Looks good logically. Will mark this as blocked until we have an associated otr-web PR.

hburn7 (dismissed):

PR #225, opened 2024-03-28
Add documentation for Me Controller

myssto:
Adds xml documentations for all endpoints and objects used in the Me Controller

### DIFF
--- a/API/DTOs/PlayerRatingChartDataPointDTO.cs
+++ b/API/DTOs/PlayerRatingChartDataPointDTO.cs
@@ -1,17 +1,67 @@
 namespace API.DTOs;
 
+/// <summary>
+/// Represents a data point used to construct a rating chart for a player
+/// </summary>
 public class PlayerRatingChartDataPointDTO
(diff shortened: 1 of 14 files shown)


API/DTOs/PlayerRatingChartDataPointDTO.cs
```diff
@@ -1,17 +1,67 @@
...
+    /// </summary>
     public double RatingChange => RatingAfter - RatingBefore;
+
+    /// <summary>
+    /// Difference in volatility for the player after this match occurred
+    /// </summary>
     public double VolatilityChange => VolatilityAfter - VolatilityBefore;
+
+    /// <summary>
+    /// Denotes the type of adjustment resulted from decay
```
hburn7, line 59:
```suggestion
    /// Indicates whether this data point is from a rating change that occurred outside of a match (i.e. decay)
```

hburn7 (changes requested):
Good stuff.

Issue #191, opened 2024-03-15
Add `/users/{id}` and `/clients/{id}` PATCH endpoints


myssto:
Currently we are assigning users and clients their claims via direct database access. We need to implement endpoints to patch these via the API directly for use by admins.

We need to identify whether these should be delegated to their own controllers (`UsersController` and `ClientsController`), or if they could be reasonably integrated into a current controller (`OAuthController`).

Tasks:
- [ ] Identify controller scheme for entities `User` and `OAuthClient`
- [ ] `PATCH` endpoint for editing `User.Scopes`
- [ ] `PATCH` endpoint for editing `OAuthClient.Scopes`

hburn7:
Using a proper `UsersController` would be appropriate here.

PR #228, opened 2024-04-01
Add Users Controller

### LINKED ISSUE #113: Return submitted tournaments with verification statuses

### LINKED ISSUE #119: Add `/users/submissions:unverify` `POST` endpoint
Add ability for administrators to unverify all user submissions at once through this endpoint.

This would do the following:

- For every tournament & match the user has submitted, set the `verification_status` to `rejected`. As of writing, we do not yet support rejecting at the tournament level.

### LINKED ISSUE #205: Add `/users/{id}/clients` `GET` endpoint
Add an endpoint to a (not yet created but planned) `UsersController` that returns the `OAuthClient` information for that user.

If #146 is complete, I believe the data should be returned unencrypted.

myssto:
This should go a long way for adding admin-panel functionality. Includes some logic improvements to the `HistoryRepositoryBase`. Implementation is complete but _needs testing_. Will ping for review when that is complete.

## Breaking Changes
- **Edit** `UserDTO` (Returns from `/me/user` and `/users/{id}`)
  - **Rename** property `OsuUsername` -> `Username`
  - **Rename** property `OsuCountry` -> `Country`
  - **Rename** property `OsuPlayMode` -> `Ruleset`

## Additions
- **Add** `GET` `/users/{id}`
- **Add** `PATCH` `/users/{id}/scopes`
- **Add** `GET` `/users/{id}/submissions`
- **Add** `POST` `/users/{id}/submissions:reject`
- **Add** `GET` `/users/{id}/clients`
- **Add** `DELETE` `/users/{id}/clients/{clientId}`

Completes Task `PATCH endpoint for editing User.Scopes` of #191
Closes #205
Closes #119
Closes #113


API/Controllers/UsersController.cs
```diff
@@ -0,0 +1,149 @@
...
+
+[ApiController]
+[ApiVersion(1)]
+[Authorize(Roles = "system, admin")] // Internal access only
+[Route("api/v{version:apiVersion}/[controller]")]
+[SuppressMessage("ReSharper", "RouteTemplates.ActionRoutePrefixCanBeExtractedToControllerRoute")]
+public class UsersController(IUserService userService, IOAuthClientService clientService) : Controller
+{
+    private readonly IUserService _userService = userService;
+    private readonly IOAuthClientService _clientService = clientService;
```
hburn7, line 20: Given the variables are made accessible to the entire class with this new constructor syntax, what do you think of removing the `private readonly` fields in classes going forward? I've been doing this in my new classes.

API/Services/Implementations/UserService.cs
```diff
@@ -29,4 +36,43 @@ public class UserService(IUserRepository repository) : IUserService
...
+        if (user is null)
+        {
+            return null;
+        }
+
+        user.Scopes = scopes.ToArray();
+        await _userRepository.UpdateAsync(user);
+
+        return _mapper.Map<UserDTO>(user);
+    }
```
hburn7, line 77: I think some extra logic here is needed. Such as forcing lower case on all scopes. Consider also restricting which scopes can be added to users (i.e. client and system should never be added to a user). Also consider checking the scopes themselves, i.e. `administrator` is not a scope we use, so it shouldn't be possible to add to a user.

API/Controllers/UsersController.cs
```diff
@@ -0,0 +1,149 @@
...
+
+    /// <summary>
+    /// Rejects a user's match submissions
+    /// </summary>
+    /// <param name="id">Id of the user</param>
+    /// <response code="404">If a user does not exist</response>
+    /// <response code="200">Denotes the operation was successful</response>
+    [HttpPost("{id:int}/submissions:reject")]
+    [ProducesResponseType(StatusCodes.Status404NotFound)]
+    [ProducesResponseType(StatusCodes.Status200OK)]
```
hburn7, line 93: Status 400 is returned but is not documented

API/Controllers/UsersController.cs
```diff
@@ -0,0 +1,149 @@
...
+    /// <param name="id">Id of the user</param>
+    /// <param name="clientId">Id of the OAuth client</param>
+    /// <response code="404">If a user or client does not exist</response>
+    /// <response code="400">If the deletion was not successful</response>
+    /// <response code="200">Denotes the deletion was successful</response>
+    [HttpDelete("{id:int}/clients/{clientId:int}")]
+    [ProducesResponseType(StatusCodes.Status404NotFound)]
+    [ProducesResponseType(StatusCodes.Status400BadRequest)]
+    [ProducesResponseType(StatusCodes.Status200OK)]
+    public async Task<IActionResult> DeleteClientAsync(int id, int clientId)
```
hburn7, line 138: Users should be able to do execute this endpoint. I think the class-level authorization needs to be removed and have each endpoint specify the authorization requirements.

API/Controllers/UsersController.cs
```diff
@@ -0,0 +1,149 @@
...
+    {
+        if (!await _clientService.ExistsAsync(clientId, id))
+        {
+            return NotFound();
+        }
+
+        return await _clientService.DeleteAsync(clientId)
+            ? BadRequest()
+            : Ok();
```
hburn7, line 147: This will return BadRequest if the operation is successful.

API/Controllers/UsersController.cs
```diff
@@ -0,0 +1,149 @@
...
+    {
+        if (!await _userService.ExistsAsync(id))
+        {
+            return NotFound();
+        }
+
+        var verifierId = HttpContext.AuthorizedUserIdentity();
+        return await _userService.RejectSubmissionsAsync(id, verifierId, (int)MatchVerificationSource.Admin)
+            ? BadRequest()
+            : Ok();
```
hburn7, line 104: This will return BadRequest if the operation is successful.

API/DTOs/MatchSubmissionStatusDTO.cs
```diff
@@ -0,0 +1,37 @@
...
+
+    /// <summary>
+    /// Lobby title of the match
+    /// </summary>
+    public string? Name { get; set; }
+
+    /// <summary>
+    /// Current verification status of the match
+    /// </summary>
+    public int? VerificationStatus { get; set; }
```
hburn7, line 26: I feel like returning the enum here is more useful

API/DTOs/UserDTO.cs
```diff
@@ -1,9 +1,42 @@
...
+
+    /// <summary>
+    /// osu! country country code of the associated player
+    /// </summary>
+    public string? OsuCountry { get; set; }
+
+    /// <summary>
+    /// Preferred osu! mode of the associated player
+    /// </summary>
+    public int? OsuPlayMode { get; set; }
```
hburn7, line 36: This would be a new user setting (not player), but I think this property can be deleted since the frontend uses cookies for this at the moment.

API/DTOs/UserDTO.cs
```diff
@@ -1,9 +1,42 @@
...
+
+    /// <summary>
+    /// osu! username of the associated player
+    /// </summary>
+    public string? OsuUsername { get; set; }
```
hburn7, line 41: I think just `Username` is fine

API/DTOs/UserDTO.cs
```diff
@@ -1,9 +1,42 @@
...
+
+    /// <summary>
+    /// osu! id of the associated player
+    /// </summary>
+    public long? OsuId { get; set; }
+
+    /// <summary>
+    /// osu! country country code of the associated player
+    /// </summary>
+    public string? OsuCountry { get; set; }
```
hburn7, line 31: I think `Country` is fine

API/Services/Interfaces/IUserService.cs
```diff
@@ -1,13 +1,46 @@
...
+    Task<IEnumerable<MatchSubmissionStatusDTO>?> GetSubmissionsAsync(int id);
+
+    /// <summary>
+    /// Rejects all match submissions of a user for the given id
+    /// </summary>
+    /// <param name="id">Id of the user</param>
+    /// <param name="verifierId">Id of the user invoking this action</param>
+    /// <param name="verificationSource">Int representation of <see cref="MatchVerificationSource"/></param>
+    /// <returns>True if successful or the user has no match submissions</returns>
+    Task<bool> RejectSubmissionsAsync(int id, int? verifierId, int? verificationSource);
```
hburn7, line 38: Rename `verifierId` to `rejecterUserId` or something a little more clear.

hburn7: I also would prefer to use the enum `MatchVerificationSource` here.

hburn7 (changes requested):
I know you said you'd ping for a review, but I found some things that could use addressing before testing. I've skipped over a lot of the new history logic at this time.

API/Controllers/UsersController.cs
myssto: That's fine with me. I've been inconsistently using either or, so let's go with using primary constructors where we can then.

hburn7 (approved):

myssto:
Updated description to include breaking changes to `UserDTO`

myssto:
PR blocked until web reflects changes

hburn7:
@AkinariHex See the "Breaking Changes" section in the PR description. Let's get this implemented asap.

bugg86 (dismissed):

hburn7 (approved):

bugg86 (dismissed):

myssto:
> Updated description to include breaking changes to `UserDTO`

Changes to `UserDTO` are now just renamed fields. `OsuPlayMode` now uses the player's default ruleset instead of always being set to `Ruleset.Standard`

AkinariHex:
@myssto does this change the keys on this object?
image

I don't see any other endpoints like `/me/user` or `/users/{id}` that you mentioned in the PR description

hburn7:
> @myssto does this change the keys on this object? image
> 
> I don't see any other endpoints like `/me/user` or `/users/{id}` that you mentioned in the PR description

image
is this what you are talking about? This was listed in the original PR note.

PR #229, opened 2024-04-03
Improve robustness of `GameAutomationChecks`

hburn7:
* Addresses issues related to checking for referees in lobby. Now, in a hypothetical 2v2 situation, referees will only be identified specifically when the team size is off by exactly 1, and only one player has a score below the minimum (currently 1000).
* In non-1v1 situations, scores less than the minimum will no longer count towards the counter for that team. For example, in a legal 2v2 (4 players in lobby), if one player on red gets <= 1000 score, that score will be removed. The check will see that there is a 1v2 for legal scores and will fail automated checks.

New tests have been added to cover the more complex scenarios for earning <=1000 score and to actually make sure we flag referees accurately. For example, new tests cover the fact that a referee should only be considered when there are *too many* players for an expected tournament team size. Previously, a 1v2 where a player earns 0 score would be flagged as a game that has a referee and would pass checks.

### DIFF
--- a/API/Osu/AutomationChecks/GameAutomationChecks.cs
+++ b/API/Osu/AutomationChecks/GameAutomationChecks.cs
@@ -1,4 +1,5 @@
 using API.Entities;
+using API.Utilities;
 
 namespace API.Osu.AutomationChecks;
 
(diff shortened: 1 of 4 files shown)


myssto (approved):
LGTM

PR #243, opened 2024-04-08
Add location service and improve search service

myssto:
Adds a location service, which is a wrapper for accessing the IUrlHelper made accessible to the DI container by AspNetCore.Mvc. This service will generate full URLs to resource locations in the API. It is meant to be a helper utility for generating `201 Created` responses and search results.

Also organizes `Program.cs` with regions for my sanity! Technically breaking changes here but nothing uses search yet so is safe to merge through.


API/DTOs/SearchResponseCollectionDTO.cs
```diff
@@ -0,0 +1,22 @@
...
+
+/// <summary>
+/// Represents a collection of search results
+/// </summary>
+public class SearchResponseCollectionDTO
+{
+    /// <summary>
+    /// A collection of search results for tournaments matching the search query
+    /// </summary>
+    public SearchResponseDTO[] Tournaments { get; set; } = null!;
```
hburn7, line 11: Any reason for using array vs ICollection / IList etc.?

myssto: > If you know the data is fixed length, and you want to micro-optimise for some very specific reason (after benchmarking), then an array may be useful.

_technically_ for data transfers arrays are marginally more efficient since they are immutable.

hburn7 (dismissed):

Issue #129, opened 2024-02-26
Convert `mode` to `ruleset` everywhere


hburn7:
Major refactor, but we need to use the term `ruleset` instead of `mode` just about everywhere. `mode` is often confused with `scoringType` and is generally just not the correct term. There could even be additional rulesets created in the future - this will help clarity significantly.

PR #247, opened 2024-04-11
Store a player's default play mode

myssto:
Small change to the `OsuPlayerDataWorker` that stores the player's default play mode when updating ranking information. This is working towards a change in how we display player profiles on web. Using this information, when visiting a player's profile on web we can show the statistics for the target player's preferred mode instead of the logged in user's preferred mode.

### DIFF
--- a/API/BackgroundWorkers/OsuPlayerDataWorker.cs
+++ b/API/BackgroundWorkers/OsuPlayerDataWorker.cs
@@ -53,7 +53,7 @@ public class OsuPlayerDataWorker(
                 {
                     // Fetch data for all game modes and update accordingly
                     var updatedOnce = false;
-                    foreach (OsuEnums.Mode gameModeEnum in Enum.GetValues<OsuEnums.Mode>())
+                    foreach (OsuEnums.Ruleset gameModeEnum in Enum.GetValues<OsuEnums.Ruleset>())
                     {
                         OsuApiUser? apiResult = await apiService.GetUserAsync(
                             player.OsuId,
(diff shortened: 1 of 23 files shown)


API/Entities/Player.cs
```diff
@@ -41,6 +42,9 @@ public class Player
     [MaxLength(4)]
     public string? Country { get; set; }
 
+    [Column("play_mode")]
+    public OsuEnums.Mode? PlayMode { get; set; }
```
hburn7, line 46: Can we call this `default_ruleset` instead?

API/Osu/Multiplayer/OsuApiService.cs
```diff
@@ -19,6 +19,9 @@ public class OsuApiUser
 
     [JsonProperty("country")]
     public string Country { get; set; } = null!;
+
+    [JsonProperty("play_mode")]
+    public GameMode PlayMode { get; set; }
```
hburn7, line 24: Here, would rather refactor `GameMode` to `Ruleset` and `PlayMode` to `Ruleset`

hburn7 (changes requested):

API/Entities/Player.cs
hburn7: And, rename `OsuEnums.Mode` to `OsuEnums.Ruleset`

hburn7:
Comments help us get a little closer towards #129

myssto:
When refactoring `OsuEnums.Mode` to `OsuEnums.Ruleset`, I opted to let Rider change whatever string versions of "mode" it found as well. We can comb through and manually change the rest in a dedicated PR.

API/BackgroundWorkers/OsuPlayerDataWorker.cs
```diff
@@ -39,57 +38,63 @@ private async Task BackgroundTask(CancellationToken cancellationToken)
...
-                    await Task.Delay(UPDATE_INTERVAL_SECONDS * 1000, cancellationToken);
+                    await Task.Delay(UpdateIntervalSeconds * 1000, cancellationToken);
                     continue;
                 }
 
                 foreach (Player? player in playersToUpdate)
                 {
-                    // Fetch ranks for all 4 game modes and update accordingly.
-                    foreach (OsuEnums.Mode gameModeEnum in Enum.GetValues<OsuEnums.Mode>())
+                    // Fetch data for all game modes and update accordingly
```
hburn7, line 54:
```suggestion
// Fetch data for all rulesets and update accordingly
```

API/Osu/Multiplayer/OsuApiService.cs
```diff
@@ -19,6 +19,9 @@ public class OsuApiUser
 
     [JsonProperty("country")]
     public string Country { get; set; } = null!;
+
+    [JsonProperty("ruleset")]
```
hburn7, line 23: I think we can actually remove all of the JsonProperty fields in this class, I don't see them being used anywhere.

hburn7 (approved):
Looks good

API/Osu/Multiplayer/OsuApiService.cs
myssto: I was wondering about those, didn't understand what they were used for so I wasn't going to touch them lol

PR #251, opened 2024-04-13
Upgrade vulnerable packages

myssto:
Yay avoiding security vulns (get this build warning off my screen)


bugg86 (approved):
hurray no more build warnings

PR #271, opened 2024-04-16
Move all `Osu/OsuEnums` into individual files at `Osu/Enums/`

myssto:
- XML documentation for all osu enums
- Moves individual enums out of static class `OsuEnums`


API/Osu/Enums/TeamType.cs
```diff
@@ -0,0 +1,29 @@
...
+{
+    /// <summary>
+    /// Free for all
+    /// </summary>
+    HeadToHead = 0,
+
+    /// <summary>
+    /// Free for all (Tag format)
+    /// </summary>
+    /// <remarks>All players play tag-team on the same beatmap</remarks>
```
hburn7, line 17:
```suggestion
    /// <remarks>All players play tag on the same beatmap</remarks>
```

hburn7 (changes requested):

hburn7 (approved):

PR #288, opened 2024-05-03
Move `/api/v1/stats/ratingadjustments DELETE` into `/api/v1/stats DELETE`

hburn7:
- Remove `/api/v1/stats/ratingadjustments DELETE`
- Merge this behavior into `/api/v1/stats DELETE`

While this is a breaking change, this was only used by the old processor.


myssto (approved):
LGTM

PR #289, opened 2024-05-03
Add the ability to enable / disable the whitelist globally

### LINKED ISSUE #232: Refactor uses of string literals for JWT claims to use `OtrClaims`
For safety and refactor friendly code, we should be using the constant strings defined in `OtrClaims` instead of string literals (Implemented in #228). For example, on a controller route, instead of `[Authorize(Roles = "system")]`, instead use `[Authorize(Roles = OtrClaims.System)]`

Depends on #228

myssto:
As we move towards a beta release, it makes sense to want to have a way to enable and disable the whitelist we have activated instead of granting all users the `whitelist` claim or remove the attributes from the controllers completely. Using the new `WhitelistEnforcementMiddleware`, the behavior can be preserved for when we need to restrict access, but also disabled when we need to do so. The middleware asserts that all requests have either the `whitelist`, `admin`, or `system` claim when active. Activation of this middleware is determined through a new config value `Auth.EnforceWhitelist`.

Since this pull edits all of the controllers anyway, I decided to include the relevant changes for 232 as well.

Closes #232

### DIFF
--- a/API/Configurations/AuthConfiguration.cs
+++ b/API/Configurations/AuthConfiguration.cs
@@ -8,4 +8,6 @@ public class AuthConfiguration
 
     [Required(ErrorMessage = "ClientCallbackUrl is required!")]
     public string ClientCallbackUrl { get; set; } = string.Empty;
+
+    public bool EnforceWhitelist { get; set; }
 }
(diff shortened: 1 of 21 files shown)


hburn7:
Tested, seems to work as expected.

hburn7 (approved):
Looks good. Will update config in CI and merge.

PR #297, opened 2024-05-07
Disable lazy loading and use `AsNoTracking` for all table fetches

### LINKED ISSUE #237: `<=` bug in leaderboard filters
This filter `https://otr.stagec.xyz/leaderboards?rating=100&rating=3150` excludes players with a rating equal to `3150`. Probably a simple `<=` bug in the leaderboard query. This also happens when filtering by matches played (on the upper bound). Does not appear to happen with rank checking.

hburn7:
Closes #237

- Disables lazy loading by default - this was causing the `GetAllAsync` repository methods to return ALL associated entities by default. Perhaps this is a reason for some of the performance issues I've noticed lately.
- Use `.AsNoTracking()` for all of the `GetAllAsync` queries unless overridden.
- Remove a useless `GetAllAsync` override in `BeatmapRepository`.

Depends on:

- [x] #299

### DIFF
--- a/API/OtrContext.cs
+++ b/API/OtrContext.cs
@@ -34,7 +34,6 @@ public partial class OtrContext(
 
     protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder) =>
         optionsBuilder
-            .UseLazyLoadingProxies()
             .UseNpgsql(_configuration.Value.DefaultConnection);
 
     protected override void OnModelCreating(ModelBuilder modelBuilder)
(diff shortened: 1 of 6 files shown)


myssto (approved):
Updated the one spot I used lazy loading in. In hindsight I am not sure why I thought it was a good idea in the first place. I realize now all it really does is let us be lazier in coding while increasing the amount of database requests...

The rest LGTM

hburn7:
I'll do a large manual test of all processes just to make sure things are good. Bugs could easily get introduced through things like automated player fetching & updating. Want to make sure the change tracker is specifically enabled there.

hburn7:
Tested against `otr-web` master, all looks good.

PR #310, opened 2024-05-11
Implement strict access control to `User` resources

myssto:
Since 99% of data that should be publicly accessible is consolidated into `Player` resources, it makes sense to internalize `User` resources to strictly admins and the resource owners.

This pull accomplishes this by implementing a custom authorization policy. The policy checks the passed `id` route parameter against the user id of the incoming request. If the ids match, or if the incoming request is of an admin user or system client, the request is authorized. This allows us to maintain our usage of redirection on `/me` endpoints while still allowing users with elevated privileges access to any user resources.

### DIFF
--- a/API/Authorization/AuthorizationPolicies.cs
+++ b/API/Authorization/AuthorizationPolicies.cs
@@ -1,0 +1,15 @@
+using API.Utilities;
+
+namespace API.Authorization;
+
+/// <summary>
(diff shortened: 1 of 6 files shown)


hburn7 (approved):
Tested locally, I don't see any issues. Didn't do the most thorough of testing but I was able to access things using the system client and all web resources.

PR #312, opened 2024-05-12
Use pagination for `GET` `/matches`

myssto:
Removes outdated methods used by the processor for requesting match data. Introduces a new endpoint for bulk-requesting match data that supports pagination.

# New Endpoints
- `GET` `/matches`
  - Returns match data in bulk
  - Pagination enforced
  - Results ordered by Id (primary key)
  - Query parameters
    - `limit` Optional, controls the amount of matches returned. 
      - `Default` 100
      - `Constraints` Minimum 1, Maximum 5000
    - `page` Optional, controls which block of size `limit` to return.
      - `Default` 1
      - `Constraints` Minimum 1
    - `filterUnverified` Optional, if unverified matches should be excluded from results. Default: True, Constraints: Requires admin or system permission if false
  - Returns a new `PagedResultDTO`
    - ```json
      {
        "next": "Nullable: Link to the next page of results",
        "previous": "Nullable: Link to the previous page of results",
        "count": 100,
        "results": []
      }
       ```
  - Example (Gets first page of 100 matches)
    - Request
      ```js
       fetch("otr.stagec.xyz/api/v1/matches?limit=100&page=1",
         {
           method: "GET",
           headers: new Headers().append("Authorization", "Bearer {}")
         });
      ```
     - Result
       ```json
       {
         "next": "otr.stagec.xyz/api/v1/matches?limit=100&page=2&filterUnverified=True",
         "previous": null,
         "count": 100,
         "results": []
       }
       ```
# Breaking Changes
## Endpoint Changes
- Remove `GET` `/matches/ids`
- Remove `GET` `/matches/id-mapping`
- Remove `POST` `/matches/convert`

## Schema Changes
- `MatchDTO` (Returns from `GET` `/matches`, etc)
  - Rename `mode` to `ruleset`
  - ```json
     {
       "id": 5838,
        "matchId": 94147364,
        "name": "BOP2: (A secret Fanclub) vs (SlavMiku)",
        "ruleset": 0,
        "startTime": "2021-11-19T08:48:31-06:00",
        "endTime": "2021-11-19T09:45:42-06:00",
        "verificationStatus": 0,
        "games": []
     }
    ```

Closes #308

PR #318, opened 2024-05-19
Fix returning invalid data in some cases

hburn7:
Depends on:

- [x] #312

Fixes some bugs relating to how data is returned when we fetch match data via pagination. Also disambiguates test functionality by introducing concrete `MatchScore` lists for games instead of just appending to those lists. Removed a lot of the logic around referee detection, instead we rely on the `MatchScore` itself being marked as `!IsValid` by a previous step in the automation checks flow.

### DIFF
--- a/API/BackgroundWorkers/OsuMatchDataWorker.cs
+++ b/API/BackgroundWorkers/OsuMatchDataWorker.cs
@@ -121,14 +121,7 @@ public class OsuMatchDataWorker(
             // Score verification checks
             foreach (MatchScore score in game.MatchScores)
             {
-                if (!ScoreAutomationChecks.PassesAutomationChecks(score))
-                {
-                    score.IsValid = false;
-                }
-                else
-                {
-                    score.IsValid = true;
-                }
+                score.IsValid = ScoreAutomationChecks.PassesAutomationChecks(score);
             }
 
             GameRejectionReason? rejectionReason = GameAutomationChecks.IdentifyRejectionReason(game);
--- a/API/BackgroundWorkers/OsuMatchDataWorker.cs
+++ b/API/BackgroundWorkers/OsuMatchDataWorker.cs
@@ -123,18 +123,11 @@ public class OsuMatchDataWorker(
             {
                 if (!ScoreAutomationChecks.PassesAutomationChecks(score))
                 {
-                    if (score.IsValid != false)
-                    {
-                        // Avoid unnecessary db calls
-                        score.IsValid = false;
-                    }
+                    score.IsValid = false;
                 }
                 else
                 {
-                    if (score.IsValid != true)
-                    {
-                        score.IsValid = true;
-                    }
+                    score.IsValid = true;
                 }
             }
 
--- a/API/BackgroundWorkers/OsuMatchDataWorker.cs
+++ b/API/BackgroundWorkers/OsuMatchDataWorker.cs
@@ -95,22 +95,20 @@ public class OsuMatchDataWorker(
         IGamesRepository gamesRepository
     )
     {
-        logger.LogInformation("Performing automated checks on match {Match}", match.MatchId);
         // Match verification checks
-        if (!MatchAutomationChecks.PassesAllChecks(match))
+        if (MatchAutomationChecks.PassesAllChecks(match))
+        {
+            match.VerificationStatus = match.VerifierUserId is not null ? MatchVerificationStatus.Verified : MatchVerificationStatus.PreVerified;
+            match.VerificationSource = MatchVerificationSource.System;
+            match.VerificationInfo = null;
+        }
+        else
         {
             match.VerificationStatus = MatchVerificationStatus.Rejected;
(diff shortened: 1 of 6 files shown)


API/BackgroundWorkers/OsuMatchDataWorker.cs
```diff
@@ -125,30 +123,11 @@ IGamesRepository gamesRepository
             {
                 if (!ScoreAutomationChecks.PassesAutomationChecks(score))
```
myssto, line 124:
```suggestion
                score.IsValid = ScoreAutomationChecks.PassesAutomationChecks(score)
```
If we don't do anything but assign `score.IsValid`, we can just use the bool returned from it instead of using an if.

API/Osu/AutomationChecks/GameAutomationChecks.cs
```diff
@@ -79,122 +80,70 @@ public static bool PassesTeamSizeCheck(Game game)
...
-                    refereePresent
+                    game.GameId
                 );
             }
 
             return satisfiesOneVersusOne;
         }
 
-        // teamSize is greater than 1 or is TeamVS
+        var countRed = validScores.Count(s => s is { Team: (int)Team.Red });
```
myssto, line 99:
```suggestion
        var countRed = validScores.Count(s => s.Team == Team.Red);
```
Don't see the need to use an object comparison here

API/Repositories/Implementations/MatchesRepository.cs
```diff
@@ -352,7 +341,8 @@ private IQueryable<Match> MatchBaseQuery(bool filterInvalidMatches)
                 .Matches.Include(x => x.Games)
                 .ThenInclude(x => x.MatchScores)
                 .Include(x => x.Games)
-                .ThenInclude(x => x.Beatmap);
+                .ThenInclude(x => x.Beatmap)
+                .OrderBy(x => x.StartTime);
```
myssto, line 345: The `GetAsync` method overrides this by ordering by primary key. Does this still work for the processor?

API/Repositories/Implementations/MatchesRepository.cs
```diff
@@ -41,17 +36,11 @@ public async Task InvalidateCacheEntriesAsync()
     [SuppressMessage("ReSharper.DPA", "DPA0007: Large number of DB records")]
     public async Task<IEnumerable<Match>> GetAsync(int limit, int page, bool filterUnverified = true)
```
myssto, line 37: Can edit this function to arrow return the query result like
```cs
public async Task<IEnumerable<Match>> GetAsync(int limit, int page, bool filterUnverified = true) =>
    await MatchBaseQuery(filterUnverified)
        .Include(m => m.Tournament)
        ...
```

myssto (changes requested):

API/Osu/AutomationChecks/GameAutomationChecks.cs
hburn7: Had it before to cover multiple props. Fixed.

API/Repositories/Implementations/MatchesRepository.cs
hburn7: The processor sorts by start time anyway I think, seems fine.

hburn7: Going to remove this orderby though, we shouldn't do that here anyway.

API/Osu/AutomationChecks/GameAutomationChecks.cs
```diff
@@ -221,11 +170,11 @@ public static bool PassesRulesetCheck(Game game)
         }
 
         s_logger.Information(
-            "{Prefix} Tournament {TournamentId} has a game mode that differs from game, can't verify game {GameId} [Tournament: Mode={TMode} | Game: Mode={GMode}",
+            "{Prefix} Tournament {TournamentId} has a game mode that differs from game, can't verify game {GameId} [Tournament: Mode={TMode} | Game: Mode={GMode}]",
```
myssto, line 173:
```suggestion
            "{Prefix} Tournament {TournamentId} has a game mode that differs from game, can't verify game {GameId} [Tournament: Ruleset={TMode} | Game: Ruleset={GMode}]",
```

API/Osu/AutomationChecks/GameAutomationChecks.cs
```diff
@@ -221,11 +170,11 @@ public static bool PassesRulesetCheck(Game game)
...
             LogPrefix,
             tournament.Id,
             game.GameId,
-            tournament.Mode,
+            (Ruleset)tournament.Mode,
             game.Ruleset
```
myssto, line 178:
```suggestion
            game.Ruleset.ToString()
```
Optional, think logs would look nicer but it's w/e

myssto (changes requested):
Functionally looks good. Not familiar with a lot of the automation checks flows so I'll leave the testing to you. Good to merge after this

PR #320, opened 2024-05-19
Rename `MatchWinRecord` properties to match new table schema

hburn7:


API/DTOs/MatchWinRecordDTO.cs
```diff
@@ -7,10 +7,10 @@ namespace API.DTOs;
 public class MatchWinRecordDTO
```
myssto, line 7: Would be a great time for some documentation :D

myssto (changes requested):

PR #321, opened 2024-05-20
Use recent tournament performances over worst performances

### LINKED ISSUE #316: Change worst tournament performance to recent tournament performance
Discussed in https://github.com/osu-tournament-rating/otr-web/discussions/186

Instead of displaying the 'worst' stats for a player's tournament history, display the most recent tournament performance (by average match cost) instead. Order by the most recent match for that tournament (e.g. first item is the most recent).

myssto:
Changes the tournament average performance calculations to only return results ordered by recent date, or best match cost.

# Breaking Changes

## Schema Changes
- `PlayerTournamentStatsDTO` (Returns from `/stats/{id}` as `tournamentStats`)
  - Renamed `worstPerformances` to `recentPerformances`
  -  ```json
       {
           "teamSizeCounts": { },
           "bestPerformances": [],
           "recentPerformances": []
       }
      ```
- `PlayerTournamentMatchCostDTO` (Structure of `PlayerTournamentStatsDTO.bestPerformances` and `PlayerTournamentStatsDTO.recentPerformances`)
  - Renamed `mode` to `ruleset`
  -  ```json
       {
           "playerId": 0,
           "tournamentId": 0,
           "tournamentName": "",
           "tournamentAcronym": "",
           "ruleset": 0,
           "matchCost": 0.0,
       }
      ```

Closes #316


API/DTOs/TournamentStatsDTO.cs
```diff
@@ -13,12 +16,12 @@ public class PlayerTournamentStatsDTO
     /// <summary>
     /// List of best tournament performances for the player
     /// </summary>
-    public IEnumerable<PlayerTournamentMatchCostDTO> BestPerformances { get; set; } =
+    public IEnumerable<PlayerTournamentMatchCostDTO> BestPerformances { get; init; } =
```
hburn7, line 19: Does init play nice with EF Core? Should we be using this everywhere? Otherwise, I'd stick to `set;` for consistency with other DTOs.

API/Repositories/Implementations/TournamentsRepository.cs
```diff
@@ -71,18 +71,19 @@ await QueryForParticipation(playerId, mode, dateMin, dateMax)
...
 
     public async Task<IEnumerable<PlayerTournamentMatchCostDTO>> GetPerformancesAsync(int playerId,
-        int mode,
+        Ruleset ruleset,
         DateTime dateMin,
         DateTime dateMax,
-        int count,
-        bool bestPerformances)
+        int limit = 5,
+        bool bestPerformances = false)
```
hburn7, line 78: I'd consider using an enum here so we can support multiple cases in the future. Something like this:

```cs
public enum PerformanceResult
{
	Best,
	Worst,
	Recent
}
```

API/Repositories/Implementations/TournamentsRepository.cs
```diff
@@ -71,18 +71,19 @@ await QueryForParticipation(playerId, mode, dateMin, dateMax)
...
     {
         IQueryable<PlayerTournamentMatchCostDTO> query =
-            QueryForParticipation(playerId, mode, dateMin, dateMax)
+            QueryForParticipation(playerId, (int)ruleset, dateMin, dateMax)
+            .OrderByDescending(t => t.Matches.Max(m => m.EndTime))
             .Select(t => new PlayerTournamentMatchCostDTO()
             {
                 PlayerId = playerId,
-                Mode = mode,
+                Mode = (int)ruleset,
```
hburn7, line 86: May as well rename `Mode` to `Ruleset` here while we're at it.

API/Repositories/Interfaces/ITournamentsRepository.cs
```diff
@@ -37,20 +38,22 @@ DateTime dateMax
     );
 
     /// <summary>
-    /// Returns a list of best or worst tournament performances for a player
+    /// Returns a list of tournament performances ordered by recent date
     /// </summary>
     /// <param name="playerId">Id (primary key) of target player</param>
-    /// <param name="mode">Ruleset</param>
+    /// <param name="ruleset">Ruleset to filter for</param>
```
hburn7, line 44: `The ruleset of the tournament`

API/Services/Implementations/PlayerStatsService.cs
```diff
@@ -335,36 +336,32 @@ DateTime dateMax
 
         IEnumerable<PlayerTournamentMatchCostDTO> bestPerformances = await _tournamentsRepository.GetPerformancesAsync(
             playerId,
-            mode,
+            (Ruleset)mode,
```
hburn7, line 339: Rename `mode` to `ruleset` here

hburn7 (changes requested):
Still need to test locally, but I'll do that after these notes are addressed. Looking forward to seeing this get merged.

PR #322, opened 2024-05-21
Allow queries for player stats to use optional ruleset

### LINKED ISSUE #238: Space should be encoded as `%20` in API calls
When looking for `[ MILK_Jiang]`, the following occurs:
```
Client with identity 20 on scheme http GET staging.otr.stagec.xyz/api/v1/players/%5B_MILK_Jiang%5D/info with body 'User with key [_MILK_Jiang] does not exist...' returned status 404
```

The first space is replaced with an underscore when manually typing into the url. Not a super high priority but something we should be aware of.

myssto:
This PR addresses two problems that exist between the frontend and the API:
- Ambiguity for a player existing, but not having data for a ruleset
- Not being able to request stats for a player's default ruleset

To address the first problem, some changes were made to the way stat generation works. Now, if a request to `/stats/{key}` specifies a ruleset but the player has no data, instead of returning a `404 Not Found`, it will return a `PlayerStatsDTO` with *only* the `playerInfo` populated. The endpoint will now correctly return a `404 Not Found` only if no player exists for the given key.

For the second problem, we are able to leverage the new settings storage to determine a player's preferred ruleset without having to specify it in the query. Now, if a request is made to `/stats/{key}` without specifying a ruleset, stats for the player's default ruleset will be returned. This allows us to mimic the functionality of the osu! website where navigating to a user page will display their preferred ruleset by default but still allowing the viewing user to switch to different rulesets if they wish.

As a bonus, this fixes most of the issues that existed API side with versatile search. Namely, issues with spaces (#238) and not being able to search for players by osu! id.

# Breaking Changes

## Endpoint Changes
- Removed `/stats/{id}`
  - Functionality now combined with `/stats/{key}` using versatile search
- Renamed `/stats/{username}` to `/stats/{key}`
  - Technically not breaking, but functionally allows `key` to be parsed as `playerId`, `osuId`, or `username`
  - Query parameters
    - Removed `?comparerId`
    - Renamed `?mode` to `?ruleset`
      - Optional, If not provided the player's default is used (In the rare case where there is no default, standard is used as a fallback)
  - Example
    - Get stats for Stage for osu!std (his default ruleset is standard)
      - Request
        ```js
         // Result will be the same for "/stats/440" (player id) or "/stats/8191845" (osu! id)
         fetch("otr.stagec.xyz/api/v1/stats/stage",
           {
             method: "GET",
             headers: new Headers().append("Authorization", "Bearer {}")
           });
        ```
      - Response (All stat fields populated)
         ```json
         {
           "playerInfo": { "id": 440, "osuId": 8191845, "username": "Stage", "country": "US" },
           "ruleset": 0,
           "baseStats": { },
           "matchStats": { },
           "modStats": { },
           "tournamentStats": { },
           "frequentTeammates": { },
           "frequentOpponents": { },
           "ratingChart": { }
         }
          ```
    - Get stats for Stage for osu!catch (he has no data for catch)
      - Request
        ```js
         // Result will be the same for "/stats/440" (player id) or "/stats/8191845" (osu! id)
         fetch("otr.stagec.xyz/api/v1/stats/stage?ruleset=2",
           {
             method: "GET",
             headers: new Headers().append("Authorization", "Bearer {}")
           });
        ```
      - Response (Still returns same schema, but all stat fields null)
         ```json
         {
           "playerInfo": { "id": 440, "osuId": 8191845, "username": "Stage", "country": "US" },
           "ruleset": 2,
           "baseStats": null,
           "matchStats": null,
           "modStats": null,
           "tournamentStats": null,
           "frequentTeammates": null,
           "frequentOpponents": null,
           "ratingChart": null
         }
          ```

## Schema Changes
- `PlayerStatsDTO` (returns from `/stats/{key}`)
  - Added `int` `ruleset`
  - All fields besides `playerInfo` and `ruleset` are now **nullable**
  - ```json
       {
         "playerInfo": { },
         "ruleset": 0,
         "baseStats": { },
         "matchStats": { },
         "modStats": { },
         "tournamentStats": { },
         "frequentTeammates": { },
         "frequentOpponents": { },
         "ratingChart": { }
       }
    ```

Closes #238


API/Repositories/Implementations/PlayerRepository.cs
```diff
@@ -56,22 +56,44 @@ public async Task<IEnumerable<Player>> GetAsync(bool eagerLoad)
...
-        if (players.Count > 1)
+        // Check for the osu id
+        if (long.TryParse(key, out var longValue))
         {
-            throw new Exception("More than one player was found.");
+            return await query.FirstOrDefaultAsync(p => p.OsuId == longValue);
         }
 
-        return players.Count == 0 ? null : players.First();
+        return await query.WhereUsername(key, false).FirstOrDefaultAsync();
```
hburn7, line 96: I would delete the if statement on 77 since this is the default anyway.

API/Repositories/Interfaces/IUserRepository.cs
```diff
@@ -15,6 +15,12 @@ public interface IUserRepository : IRepository<User>
     /// </summary>
     Task<User> GetByPlayerIdOrCreateAsync(int playerId);
 
+    /// <summary>
+    /// Gets a user's player id for the given id
```
hburn7, line 19: `Gets a user's player id`

API/Repositories/Interfaces/IUserRepository.cs
```diff
@@ -15,6 +15,12 @@ public interface IUserRepository : IRepository<User>
...
+    /// </summary>
+    /// <returns> A player id, or null if a user was not found or the user has no player entry </returns>
+    Task<int?> GetPlayerIdAsync(int id);
```
hburn7, line 22: Document `id`

API/Services/Interfaces/IPlayerService.cs
```diff
@@ -1,22 +1,14 @@
...
 
 public interface IPlayerService
 {
-    /// <summary>
-    /// Denotes whether a player for the given id exists
-    /// </summary>
-    Task<bool> ExistsAsync(int id);
-
     Task<IEnumerable<PlayerDTO>> GetAllAsync();
+
```
hburn7, line 8: I think we only use blank lines in cases where there's documentation, unless this was done by the formatter?

API/Repositories/Implementations/PlayerRepository.cs
hburn7, line 96: I would remove the if statement on line 77 if this is the default anyway.

hburn7 (changes requested):
Looks good for the most part, still need to test on local. Will do that after comments are addressed.

API/Repositories/Implementations/PlayerRepository.cs
myssto: Yea I get the itch to do this every time I look at it but I was going to save it for a pull dedicated to code quality

hburn7: that's fine

API/Services/Interfaces/IPlayerService.cs
myssto: Formatter I think :P Either way, I'll be doing some code quality stuff on the repos soon that will include docs 👍

hburn7 (approved):

PR #326, opened 2024-05-29
Convert `mode` to `ruleset` everywhere

foxtail286:
Might have missed something, should be a near-total refactor though


hburn7:
Thanks for attempting this, unfortunately we are in the middle of a huge restructuring. You can see this on one of our branches (think it has rewrite in the name, just check our branch list and you'll see).

I don't think these changes will see the light of day unfortunately. Please let us know in advance if you plan to touch so many files in the future.

foxtail286:
Alright, thank you. Will do 👍

Issue #371, opened 2024-08-31
Admin panel support


hburn7:
# Admin panel

- [x] #378
- [x] #379
- [x] Expand `/tournaments` `GET` to include `VerificationStatus`, `ProcessingStatus`
- [x] #400
- [x] #401

hburn7:
Reminder to flesh out this issue further.

myssto:
Picking this up now while we wait for processor to catch up

PR #399, opened 2024-09-15
Include ProcessingStatus & VerificationStatus in TournamentDTO

hburn7:
Satisfies `Expand /tournaments GET to include VerificationStatus, ProcessingStatus` portion of #371


API/DTOs/TournamentDTO.cs
```diff
@@ -33,6 +35,16 @@ public class TournamentDTO
     /// <example>For a 10,000-50,000 tournament, this value would be 10,000</example>
     public int RankRangeLowerBound { get; set; }
 
+    /// <summary>
+    ///     The current state of verification
```
myssto, line 39: Why the extra spaces here?

myssto (changes requested):

API/DTOs/TournamentDTO.cs
hburn7: Not sure what happened there

myssto (approved):

PR #405, opened 2024-09-16
Tournament verification query param

### LINKED ISSUE #400: Expand `/tournaments/{id}` `GET` to add support for including unverified data
- [x] Add `?unfiltered=true`  to `/tournaments/{id}` to include all match data, regardless of verification status
  - [x] If `?unfiltered=true` is specified, return **ALL** data for the tournament, including all child navigations.
  - [x] If `?unfiltered=false` is specified, always return fully verified data with limited navigations.

hburn7:
Closes #400 

If `unfiltered=true` is specified, all child navigations are returned as defined in `TournamentsBaseQuery()`. If `unfiltered=false` is specified, returns all child navigations with verified data.


alixdotsh (approved):
everything looks good on here. i think the only thing that was questionable was some of the formatting changes but i assume its IDE auto formatting or `.editorconfig`

Database/Repositories/Implementations/TournamentsRepository.cs
```diff
@@ -17,6 +18,18 @@ public class TournamentsRepository(OtrContext context) : RepositoryBase<Tourname
...
 
+    public async Task<Tournament?> GetVerifiedAsync(int id) =>
+        await _context.Tournaments
+            .AsSplitQuery()
+            .Include(t => t.Matches.Where(m => m.VerificationStatus == VerificationStatus.Verified))
+            .ThenInclude(m => m.Games.Where(g => g.VerificationStatus == VerificationStatus.Verified))
+            .ThenInclude(g => g.Beatmap)
+            .Include(t => t.Matches.Where(m => m.VerificationStatus == VerificationStatus.Verified))
+            .ThenInclude(m => m.Games.Where(g => g.VerificationStatus == VerificationStatus.Verified))
+            .ThenInclude(g => g.Scores.Where(gs => gs.VerificationStatus == VerificationStatus.Verified))
```
myssto, line 29: This should probably also filter for processing status

myssto (changes requested):

myssto (approved):

Issue #351, opened 2024-08-27
Crash GET request on specified match endpoint


AkinariHex:
When calling a GET request to get informations about a specified match it crashes.

API endpoint: `/matches/{id}`
Tested in branch: `feature/data-processor-rewrite`

```console
[2024-08-27 17:27:36.464 INF] Client on scheme http requests GET localhost:5075/api/v1/matches/72437 with body ''
[2024-08-27 17:27:36.488 ERR] An unhandled exception has occurred while executing the request.
System.InvalidOperationException: The expression 'x.Games.AsQueryable().WhereFiltered(__filterType_0)' is invalid inside an 'Include' operation, since it does not represent a property access: 't => t.MyProperty'. To target navigations declared on derived types, use casting ('t => ((Derived)t).MyProperty') or the 'as' operator ('t => (t as Derived).MyProperty'). Collection navigation access can be filtered by composing Where, OrderBy(Descending), ThenBy(Descending), Skip or Take operations. For more information on including related data, see https://go.microsoft.com/fwlink/?LinkID=746393.
   at Microsoft.EntityFrameworkCore.Query.Internal.NavigationExpandingExpressionVisitor.<ProcessInclude>g__ExtractIncludeFilter|36_0(Expression currentExpression, Expression includeExpression)
   at Microsoft.EntityFrameworkCore.Query.Internal.NavigationExpandingExpressionVisitor.ProcessInclude(NavigationExpansionExpression source, Expression expression, Boolean thenInclude, Boolean setLoaded)
   at Microsoft.EntityFrameworkCore.Query.Internal.NavigationExpandingExpressionVisitor.VisitMethodCall(MethodCallExpression methodCallExpression)
   at Microsoft.EntityFrameworkCore.Query.Internal.NavigationExpandingExpressionVisitor.VisitMethodCall(MethodCallExpression methodCallExpression)
   at Microsoft.EntityFrameworkCore.Query.Internal.NavigationExpandingExpressionVisitor.VisitMethodCall(MethodCallExpression methodCallExpression)
   at Microsoft.EntityFrameworkCore.Query.Internal.NavigationExpandingExpressionVisitor.VisitMethodCall(MethodCallExpression methodCallExpression)
   at Microsoft.EntityFrameworkCore.Query.Internal.NavigationExpandingExpressionVisitor.VisitMethodCall(MethodCallExpression methodCallExpression)
[...]

PR #406, opened 2024-09-16
Add `HTTP PATCH` endpoints for `Match`, `Game`, and `GameScore`

### LINKED ISSUE #401: Add `PATCH` endpoints for admin panel support
- [x] Add `/matches` `PATCH`
  - [x] Add `/games` `PATCH` (`GamesController`)
  - [x] Add `/scores` `PATCH` (`ScoresController`)

hburn7:
Depends on:

- #351 
- [x] Test `/matches/{id} PATCH` after #351 is resolved.

Closes #401 

# Changes

- Adds `PATCH` endpoint for `Match`, `Game`, and `GameScore`.
- Adds `GamesController`
- Adds `GameScoresController`
- Adds service classes and repository methods where necessary.


API/Controllers/GameScoresController.cs
```diff
@@ -0,0 +1,57 @@
...
+
+namespace API.Controllers;
+
+[ApiController]
+[ApiVersion(1)]
+[Route("api/v{version:apiVersion}/[controller]")]
+public class GameScoresController(IGameScoresService gameScoresService) : Controller
+{
+    /// <summary>
+    ///     Amend score data
```
myssto, line 18:
```suggestion
    /// Amend score data
```

API/Controllers/GameScoresController.cs
```diff
@@ -0,0 +1,57 @@
...
+    [Authorize(Roles = OtrClaims.Admin)]
+    [ProducesResponseType(StatusCodes.Status404NotFound)]
+    [ProducesResponseType(StatusCodes.Status400BadRequest)]
+    [ProducesResponseType<string>(StatusCodes.Status400BadRequest)]
+    [ProducesResponseType<GameScoreDTO>(StatusCodes.Status200OK)]
+    [HttpPatch("{id:int}")]
+    public async Task<IActionResult> UpdateAsync(int id, [FromBody] JsonPatchDocument<GameScoreDTO> patch)
+    {
+        // Ensure target game score exists
+        GameScoreDTO? game = await gameScoresService.GetAsync(id);
```
myssto, line 34: var name

myssto: nice copy paste!

API/Controllers/MatchesController.cs
```diff
@@ -70,6 +71,45 @@ public async Task<IActionResult> GetAsync(int id)
     // TODO: Should be /player/{osuId}/matches instead.
     [HttpGet("player/{osuId:long}")]
     [Authorize(Roles = $"{OtrClaims.Admin}, {OtrClaims.System}")]
-    public async Task<ActionResult<IEnumerable<MatchDTO>>> GetMatchesAsync(long osuId, Ruleset ruleset) =>
+    public async Task<IActionResult> GetMatchesAsync(long osuId, Ruleset ruleset) =>
         Ok(await matchesService.GetAllForPlayerAsync(osuId, ruleset, DateTime.MinValue, DateTime.MaxValue));
+
+    /// <summary>
+    ///     Amend match data
```
myssto, line 78:
```suggestion
    /// Amend match data
```

API/Controllers/TournamentsController.cs
```diff
@@ -57,7 +57,8 @@ public async Task<IActionResult> CreateAsync([FromBody] TournamentSubmissionDTO
 
         if (await tournamentsService.ExistsAsync(tournamentSubmission.Name, tournamentSubmission.Ruleset))
         {
-            return BadRequest($"A tournament with name {tournamentSubmission.Name} for ruleset {tournamentSubmission.Ruleset} already exists");
+            return BadRequest(
+                $"A tournament with name {tournamentSubmission.Name} for ruleset {tournamentSubmission.Ruleset} already exists");
```
myssto, line 61: I'm going to assume this was your formatter. I wonder why it behaves differently for everyone it's so odd... Maybe we need to dedicate an issue to taking another look at the editorconfig and setting some more rules

API/DTOs/GameScoreDTO.cs
```diff
@@ -6,7 +6,7 @@ public class GameScoreDTO
 {
     public int PlayerId { get; set; }
```
myssto, line 7: Add docstrings

API/Services/Implementations/GameScoresService.cs
```diff
@@ -0,0 +1,30 @@
...
+        GameScore? existing = await gameScoresRepository.GetAsync(id);
+        if (existing is null)
+        {
+            return null;
+        }
+
+        existing.Team = score.Team;
+        existing.Score = score.Score;
+        existing.Mods = score.Mods;
+        existing.CountMiss = score.Misses;
```
myssto, line 25: This should be able to update the `VerificationStatus`, `ProcessingStatus`, and `RejectionReason` as well

API/DTOs/GameScoreDTO.cs
myssto: Also add `VerificationStatus`, `ProcessingStatus`, and `RejectionReason`

API/Services/Implementations/GamesService.cs
```diff
@@ -0,0 +1,32 @@
...
+        {
+            return null;
+        }
+
+        existing.Ruleset = game.Ruleset;
+        existing.ScoringType = game.ScoringType;
+        existing.TeamType = game.TeamType;
+        existing.Mods = game.Mods;
+        existing.StartTime = game.StartTime;
+        existing.EndTime = game.EndTime ?? existing.EndTime;
```
myssto, line 27: Same here as with `GameScoresService.UpdateAsync`, should be able to update processing enums

Database/Repositories/Implementations/GamesRepository.cs
```diff
@@ -1,17 +1,12 @@
 using System.Diagnostics.CodeAnalysis;
 using Database.Entities;
 using Database.Repositories.Interfaces;
-using Microsoft.EntityFrameworkCore;
 
 namespace Database.Repositories.Implementations;
 
-[SuppressMessage("Performance", "CA1862:Use the \'StringComparison\' method overloads to perform case-insensitive string comparisons")]
+[SuppressMessage("Performance",
+    "CA1862:Use the \'StringComparison\' method overloads to perform case-insensitive string comparisons")]
```
myssto, line 8: lmao

myssto (changes requested):
These patch endpoints should support updating the data worker related enums as well. Most of this looks good though! Nice work :)

API/Controllers/TournamentsController.cs
hburn7: Definitely the formatter, really not sure why it does this on my laptop.

API/DTOs/GameDTO.cs
```diff
@@ -39,6 +40,21 @@ public class GameDTO
     /// </summary>
     public long OsuId { get; set; }
 
+    /// <summary>
+    /// The current state of verification
+    /// </summary>
+    public VerificationStatus VerificationStatus { get; set; }
```
alixdotsh, line 46: is there a possibility to rename the system as `CurrentVerificationStatus`? so in the end it would look like `public CurrentVerificaitonStatus VerificationStatus { get; set; }`, i have a feeling this could cause some issues along the way.

The only reason I mention this is bc in my past code normally I would have to do something like this but it oculd not be the same name twice.

myssto: That would involve renaming the enum so no. Personally I think it's fine as is considering there is no real confusion over the type of this property

myssto (approved):

API/DTOs/GameDTO.cs
hburn7: Agreed with Myssto, I don't see any issue with the enum & property name matching. Pretty common in C# afaik.

alixdotsh: ah im used to it not being like that in the web dev world. Outside of that everything here is fine

alixdotsh (approved):

PR #408, opened 2024-09-17
Tournament admin notes

hburn7:
Adds tournament admin notes entity

PR #409, opened 2024-09-17
Match admin notes

hburn7:
Depends on:
- #408

Adds admin notes for matches


alixdotsh (approved):
meowdy, code looks good to me 🐱 👍

Database/Entities/Match.cs
```diff
@@ -117,8 +117,17 @@ public class Match : UpdateableEntityBase, IAuditableEntity<MatchAudit>, IProces
     /// </summary>
     public ICollection<RatingAdjustment> PlayerRatingAdjustments { get; set; } = new List<RatingAdjustment>();
 
+    /// <summary>
+    /// A collection of <see cref="MatchAudit"/>s, one for each change made to this entity
+    /// </summary>
```
myssto, line 122: Technically this isn't required, it inherits the generic documentation from the interface

Database/Entities/Match.cs
```diff
@@ -117,8 +117,17 @@ public class Match : UpdateableEntityBase, IAuditableEntity<MatchAudit>, IProces
...
     public ICollection<MatchAudit> Audits { get; set; } = new List<MatchAudit>();
 
+    /// <summary>
+    /// A collection of <see cref="AdminNotes"/>s. Admins may leave notes on matches detailing
+    /// important information
+    /// </summary>
```
myssto, line 128: See above comment

myssto (approved):
Approving but see comments

PR #415, opened 2024-09-19
DataWorker fix a crash for tournaments with no matches

myssto:


hburn7 (approved):
Applying as a bandaid fix so it doesn't crash in my sleep, would like to avoid re-polling these tournaments if possible, per discord discussion

PR #425, opened 2024-09-23
Create `TAdminNoteEntity` repository base

hburn7:
Creates a base repository for all `AdminNote` entities

PR #426, opened 2024-09-23
Add `MatchAdminNoteRepository`, `MatchAdminNoteDTO` and update `MatchDTO`

hburn7:
Depends on #425

Adds `MatchAdminNoteRepository`, `MatchAdminNoteDTO` and updates `MatchDTO` to include a nullable collection of admin notes.

@myssto, is it safe to use `opt.Ignore()` for this in the `MapperProfile`? I think we need to map this manually if we do include a `UserDTO` inside of the `MatchAdminNoteDTO`. Let me know your thoughts on all of that though.


hburn7:
Closing as we need to make these repositories generic.

PR #431, opened 2024-09-27
Access layer for admin notes

myssto:
Depends on #425

Adds an API access layer to admin notes through a `AdminNoteService` and `AdminNoteDTO`


hburn7 (approved):

hburn7:
Part of #371

Issue #379, opened 2024-09-03
Create a `POST`, `PATCH`, and `DELETE` endpoint for each admin note entity and establish access layers


myssto:
Depends on #378

Admin notes need API routes for CRUD actions. They should be routed like: 

`POST` `/{entity}/{id}/notes`
`GET` `/{entity}/{id}/notes`
`PATCH` `/{entity}/{id}/notes/{noteId}`
`DELETE` `/{entity}/{id}/notes/{noteId}`

Admin notes should be included on the entity's existing DTO for admin users only.

# Tasks

Implement the four endpoints (and `AdminNote` repository / service layers) for these entities:

**Base classes**
- [x] `AdminNoteRepositoryBase` #431
- [x] `AdminNoteService` #431

- **Tournament** #433
  - [x] `POST`
  - [x] `GET`
  - [x] `PATCH`
  - [x] `DELETE`

- **Match** #456
  - [x] `POST`
  - [x] `GET`
  - [x] `PATCH`
  - [x] `DELETE`

- **Game** #457 
  - [x] `POST`
  - [x] `GET`
  - [x] `PATCH`
  - [x] `DELETE`

- **GameScore** #458
  - [x] `POST`
  - [x] `GET`
  - [x] `PATCH`
  - [x] `DELETE`

- **Player** #459
  - [x] `POST`
  - [x] `GET`
  - [x] `PATCH`
  - [x] `DELETE`

hburn7:
Removed `OAuthClient` and `Player` from the scope of this issue as they are not particularly relevant right now. (i.e., I don't see a use for admin notes in either of these entities currently). We also don't have a UI for that yet.

hburn7:
Closed once #456, #457, #458, and #459 are merged.

PR #433, opened 2024-09-27
Admin notes/tournament endpoints

myssto:
Depends on #431
Completes `Tournament` task of #379

I wanted to experiment with a little organization in this one. Since some of the controllers are going to start getting quite large, I think it would be nice if we could branch off certain groups of features into their own partial. For example here, any admin feature related endpoint definitions for `/tournaments` would now exist in the `API/Controllers/TournamentsController.Admin.cs` file.


hburn7:
Did you also run into this?
> [2024-09-27 00:05:22.352 ERR] An unhandled exception has occurred while executing the request.
Swashbuckle.AspNetCore.SwaggerGen.SwaggerGeneratorException: Conflicting method/path combination "PATCH api/v1/tournaments/{id}/notes/{noteId}" for actions - API.Controllers.TournamentsController.UpdateAdminNoteAsync (API),API.Controllers.TournamentsController.DeleteAdminNoteAsync (API). Actions require a unique method/path combination for Swagger/OpenAPI 3.0. Use ConflictingActionsResolver as a workaround
   at Swashbuckle.AspNetCore.SwaggerGen.SwaggerGenerator.GenerateOperations(IEnumerable`1 apiDescriptions, SchemaRepository schemaRepository)
   at Swashbuckle.AspNetCore.SwaggerGen.SwaggerGenerator.GeneratePaths(IEnumerable`1 apiDescriptions, SchemaRepository schemaRepository)
   at Swashbuckle.AspNetCore.SwaggerGen.SwaggerGenerator.GetSwaggerDocumentWithoutFilters(String documentName, String host, String basePath)
   at Swashbuckle.AspNetCore.SwaggerGen.SwaggerGenerator.GetSwaggerAsync(String documentName, String host, String basePath)
   at Swashbuckle.AspNetCore.Swagger.SwaggerMiddleware.Invoke(HttpContext httpContext, ISwaggerProvider swaggerProvider)
   at Microsoft.AspNetCore.RateLimiting.RateLimitingMiddleware.InvokeInternal(HttpContext context, EnableRateLimitingAttribute enableRateLimitingAttribute)
   at Microsoft.AspNetCore.Authorization.AuthorizationMiddleware.Invoke(HttpContext context)
   at Microsoft.AspNetCore.Authentication.AuthenticationMiddleware.Invoke(HttpContext context)
   at API.Middlewares.RequestLoggingMiddleware.Invoke(HttpContext context) in /Users/stage/Documents/github/otr-api/API/Middlewares/RequestLoggingMiddleware.cs:line 20
   at Microsoft.AspNetCore.Diagnostics.DeveloperExceptionPageMiddlewareImpl.Invoke(HttpContext context)

myssto:
yea oops W copy paste from me. the patch and delete methods were both using a `[HttpPatch]` attribute

hburn7 (approved):
Can be merged once all endpoints / service methods etc. are tested locally.

API/Controllers/TournamentsController.Admin.cs
```diff
@@ -0,0 +1,162 @@
...
+    /// <response code="404">If a tournament matching the given id does not exist</response>
+    /// <response code="400">If the authorized user does not exist</response>
+    /// <response code="200">Returns the created admin note</response>
+    [HttpPost("{id:int}/notes")]
+    [Authorize(Roles = $"{OtrClaims.Admin}")]
+    [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+    [ProducesResponseType(StatusCodes.Status404NotFound)]
+    [ProducesResponseType(StatusCodes.Status400BadRequest)]
+    [ProducesResponseType<AdminNoteDTO>(StatusCodes.Status200OK)]
+    public async Task<IActionResult> CreateAdminNoteAsync(int id, [FromBody] string note)
```
hburn7, line 26: In testing, I hard-coded my user id (20). The `adminUsername` field returned an empty string. From what I can see in the current code, there isn't a way for the `CreateAsync()` method to link back to a user. 
CleanShot 2024-09-28 at 11 48 20@2x

hburn7 (changes requested):

API/Controllers/TournamentsController.Admin.cs
hburn7: I see, I guess I was expecting the returned entity to have a populated user object? I imagine the following:
- Admin creates note
- Note is displayed on entity (immediately, no additional GET call required).

I think it makes sense to have a populated `AdminUser` here then.

myssto: We can load the navigation by force after the entity is created or we can simply save it to the database, and then immediately fetch it from the database using `AdminNoteRepository.GetAsync<T>(createdNote.Id)` since that method includes the required navigations. I think the later is probably best.

hburn7: Yeah I agree with you. So, basically just adding a call above the `return mapper.Map` which calls `.GetAsync` on the newly created entity to load the user navigation. If so that sounds good.

hburn7:
Updated to configure relationship between `User` and `TournamentAdminNote`.

Issue #336, opened 2024-08-10
Add missing information for tournaments page


hburn7:
- [x] Pagination for `/tournaments` https://github.com/osu-tournament-rating/otr-api/pull/436
- [x] Include `UserDTO` `submitter` on `/tournaments` `GET`
- [x] Include `UserDTO` `submitter` on `/tournaments/{id}` `GET`
- [x] Tournament-specific search
- [x] Ensure `startDate` and `endDate` are returned. These will be the dates of the first and last matches (start times of these matches)

hburn7:
As for tournament-specific search, for now just use the existing search endpoint and only use the results from the `tournaments` field. @AkinariHex

PR #436, opened 2024-09-28
Add tournament pagination

hburn7:
Part of #336 

Adds tournament pagination by changing how `/tournaments GET` works. The endpoint now accepts an optional `TournamentRequestQueryDTO` (very similar to the leaderboard implementation). This allows us to add query parameters for tournament pagination and other filters (e.g. sorting) which can come later.

Also changes the query parameter `unfiltered` to `verified` (thus inverting its behavior) in `/tournaments/{id} GET`.


API/Controllers/TournamentsController.cs
```diff
@@ -15,17 +15,33 @@ namespace API.Controllers;
...
     /// <summary>
-    /// List all tournaments
+    /// Get all tournaments which fit an optional request query
     /// </summary>
+    /// <param name="requestQuery">The optional request query filter</param>
     /// <remarks>Will not include match data</remarks>
-    /// <response code="200">Returns all tournaments</response>
+    /// <response code="200">Returns all tournaments which fit the request query</response>
+    /// <response code="401">If a request is made to fetch all tournaments without
+    /// the system claim</response>
```
myssto, line 24: Move opening and closing xml tags to their own line

API/Controllers/TournamentsController.cs
```diff
@@ -15,17 +15,33 @@ namespace API.Controllers;
...
+    public async Task<IActionResult> ListAsync([FromQuery] TournamentRequestQueryDTO? requestQuery)
     {
-        IEnumerable<TournamentDTO> result = await tournamentsService.ListAsync();
-        return Ok(result);
+        if (requestQuery is null)
+        {
+            if (!HttpContext.User.IsSystem())
+            {
+                return Unauthorized();
+            }
```
myssto, line 37: What is this and why are we doing it...

API/Controllers/TournamentsController.cs
```diff
@@ -15,17 +15,33 @@ namespace API.Controllers;
...
+    /// <param name="requestQuery">The optional request query filter</param>
     /// <remarks>Will not include match data</remarks>
-    /// <response code="200">Returns all tournaments</response>
+    /// <response code="200">Returns all tournaments which fit the request query</response>
+    /// <response code="401">If a request is made to fetch all tournaments without
+    /// the system claim</response>
+    /// <returns>All tournaments if the <param name="requestQuery">requestQuery</param> is null,
+    /// else returns all tournaments which fit the query
     [HttpGet]
-    [Authorize(Roles = OtrClaims.System)]
```
myssto, line 23: If the intention was to only auth on system clients then why did you remove the attribute

API/Controllers/TournamentsController.cs
```diff
@@ -75,21 +91,22 @@ public async Task<IActionResult> CreateAsync([FromBody] TournamentSubmissionDTO
     /// Get a tournament
     /// </summary>
     /// <param name="id">Tournament id</param>
-    /// <param name="unfiltered">If true, includes all match data, regardless of verification status.
-    /// Also includes all child navigations if true.
-    /// Default false (strictly verified data with limited navigation properties)
+    /// <param name="verified">If true, specifically includes verified match data. If false,
```
myssto, line 94:
```suggestion
    /// <param name="verified">
    /// If true, specifically includes verified match data. If false,
```
When the xml exceeds 1 line in length I would prefer to expand and keep the opening tag in it's own line

API/Controllers/TournamentsController.cs
myssto: This should not be a system-only endpoint it should be a user/client endpoint

myssto: 401 Does not need to be documented here, it is implicit because of the authorize attribute

API/DTOs/TournamentRequestQueryDTO.cs
```diff
@@ -0,0 +1,24 @@
+using Database.Enums;
+
+namespace API.DTOs;
+
+// Can be expanded to support filters like leaderboards if needed
+public class TournamentRequestQueryDTO
```
myssto, line 6: Add summary

API/DTOs/TournamentRequestQueryDTO.cs
```diff
@@ -0,0 +1,24 @@
...
+    /// </summary>
+    public int PageSize { get; set; } = 20;
+    /// <summary>
+    /// Whether the tournaments must be verified
+    /// </summary>
+    public bool Verified { get; set; } = true;
+    /// <summary>
+    /// An optional ruleset to filter by
+    /// </summary>
+    public Ruleset? Ruleset { get; set; }
```
myssto, line 23:
```suggestion
    /// <summary>
    /// The page
    /// </summary>
    public int Page { get; set; }
    
    /// <summary>
    /// The size of the page
    /// </summary>
    public int PageSize { get; set; } = 20;
    
    /// <summary>
    /// Whether the tournaments must be verified
    /// </summary>
    public bool Verified { get; set; } = true;
    
    /// <summary>
    /// An optional ruleset to filter by
    /// </summary>
    public Ruleset? Ruleset { get; set; }
```
Add new lines between properties please

API/Services/Implementations/TournamentsService.cs
```diff
@@ -86,4 +96,17 @@ public async Task<int> CountPlayedAsync(
...
+        if (requestQuery.Page < 1)
+        {
+            throw new ArgumentOutOfRangeException(nameof(requestQuery.Page), "Page must be greater than 0");
+        }
+
+        if (requestQuery.PageSize < 1)
+        {
+            throw new ArgumentOutOfRangeException(nameof(requestQuery.PageSize), "PageSize must be greater than 0");
+        }
+    }
```
myssto, line 111: Use component model data annotation attributes on the DTO instead of this

myssto (changes requested):

API/Controllers/TournamentsController.cs
hburn7: So, the intention here is to have a public endpoint (this should even be `AllowAnonymous` now that I think about it) that returns paginated tournaments. However, at least in my head, in order to retain the "get all" functionality, I opted to have `requestQuery` be optional. If the `requestQuery` is null and the authorized user is a system user, then we return all tournament data at once.

I did this because there is a conflict on `/tournaments GET` - I am unsure what the best solution is for having both a "filtered" get and an actual "get me everything" endpoint (at least from a naming perspective).

myssto: I just don't see why we need a "get it all at once" endpoint available at all, it sounds like a terrible idea. The amount of data you are loading into memory and then transferring over network in one transaction by doing that is absurd. The processor doesn't even interact with the API anymore (not to mention it doesnt pull this data to begin with) and that is the only use case I can think of. Pagination should be the only way we support getting batches of entities IMO

myssto: `/GET` `/tournaments` should be a means by which you get a batch of information about tournaments themselves. If one wants to get the entire tournament's store of match data, they should take a result from `/GET` `/tournaments` and request it from `/GET` `/tournaments/{id}`

We **SHOULD**:
- Allow granular filtering (by name, ruleset, verification status, processing status, submitter)
- Include compact DTOs for the submitter and verifier

We **SHOULDNT**:
- Include deep navigational data (matches, games, scores, beatmaps, etc)

hburn7: Yes all of these things are what this change tries to accomplish. Should both the GET all and GET (with filter) live under the same endpoint (how it is now)? If not, what should the naming scheme be?

hburn7: > I just don't see why we need a "get it all at once" endpoint available at all, it sounds like a terrible idea. The amount of data you are loading into memory and then transferring over network in one transaction by doing that is absurd. The processor doesn't even interact with the API anymore (not to mention it doesnt pull this data to begin with) and that is the only use case I can think of. Pagination should be the only way we support getting batches of entities IMO

That isn't the behavior though. This filter is for pagination. Without the filter, it acts as a standard `ListAsync`. So, if the filter is present (which it must be for any non-system web request), it will also be paginated.

hburn7: And yes because the processor doesn't use the API, I see no use case for a system-level client permission throughout the entire application. So if we want to just completely nuke that capability altogether (as I see no use case for a `/tournaments GET` returning all tournaments), then we can. I only kept this functionality as we tend to have a `/entity GET` which returns everything. 

Bigger question: Why have `ListAsync()` style endpoints at all?

myssto: > That isn't the behavior though. This filter is for pagination. Without the filter, it acts as a standard ListAsync. So, if the filter is present (which it must be for any non-system web request), it will also be paginated.

I see, I didn't really understand what you were going for but I get it. What I did for the matches endpoint was just enforce the binding of `page` and `limit` so that the endpoint is pagination-only.

> I see no use case for a system-level client permission throughout the entire application.

I agree, we can probably axe that at this point

> Bigger question: Why have ListAsync() style endpoints at all?

`ListAsync()` is still the preferred naming scheme for this style of endpoint (see [here](https://cloud.google.com/apis/design/standard_methods#list)). I just think we were at a bit of a disconnect as to what it actually meant in the first place :P

hburn7: I will enforce the bindings of Page / PageSize here and remove the system client check for returning all tournaments.

alixdotsh (approved):

myssto (approved):
Nice work :D

hburn7:
> Nice work :D

:D

PR #444, opened 2024-09-30
Refactor usage of `claim` to `role` in the context of JWT

myssto:
We inconsistently (and incorrectly in most places) use the term `claim` to describe a user / client permission that we encode into JWTs for access control. The correct term for this would be `role`.

In the `OAuthHandler` we encode `User.Scopes` and `Client.Scopes` into the JWT as claims with a `ClaimType` of `ClaimTypes.Role`. In the JWT they appear as such:

```json
{
  "role": [
     "whitelist",
     "admin",
     "user"
  ]
}
```

I know this is mostly semantics, but it helps cut confusion especially since there are spots where we do actually encode a custom `ClaimType` like rate limit overriding. This PR is just the refactoring for namespaces and naming convention. I will be updating some code for OAuth as well.


API/Authorization/AuthorizationPolicies.cs
```diff
@@ -8,7 +8,7 @@ namespace API.Authorization;
 public static class AuthorizationPolicies
 {
     /// <summary>
-    /// Policy that allows access from <see cref="OtrClaims.Admin"/> and <see cref="OtrClaims.System"/> requests,
+    /// Policy that allows access from <see cref="OtOtrClaimsdmin"/> and <see cref="OtOtrClaimsystem"/> requests,
```
hburn7, line 11: typo?

API/Authorization/OtrClaims.cs
```diff
@@ -0,0 +1,113 @@
...
+                Submitter => true,
+                Whitelist => true,
+                RateLimitOverrides => true,
+                _ => false
+            };
+        }
+    }
+
+    /// <summary>
+    /// Claim granted to users or clients to denote overrides to the default rate limit
```
hburn7, line 110: I don't know why users would ever need a rate limit override, would like to remove in the future if you also don't see a use case.

API/Handlers/Implementations/OAuthHandler.cs
```diff
@@ -248,13 +249,13 @@ private string WriteToken(SecurityTokenDescriptor tokenDescriptor)
     /// </summary>
     /// <param name="issuer">The id of the user or client this token is generated for</param>
     /// <param name="role">
-    /// The role value to encode to the JWT. This value needs to be either <see cref="OtrClaims.User"/>
-    /// or <see cref="OtrClaims.Client"/> depending on what type of entity this token is generated for
+    /// The role value to encode to the JWT. This value needs to be either <see cref="OtOtrClaimsser"/>
+    /// or <see cref="OtOtrClaimslient"/> depending on what type of entity this token is generated for
```
hburn7, line 253: typos

hburn7 (changes requested):

API/Authorization/OtrClaims.cs
myssto: I agree, this can probably be removed

hburn7 (approved):

PR #474, opened 2024-10-08
Support including mappools in tournament submission

myssto:
Closes #437
Closes #439
Closes #440

PR #475, opened 2024-10-08
Allow tournaments to be rejected at submission

### LINKED ISSUE #357: Add capability to submit a tournament with the intention of rejecting it.
Incorporate the following into the current `/tournaments POST` endpoint:

- [x] In `TournamentSubmissionDTO`, `rejectionReason` `TournamentRejectionReason`
- [x] Server-side check for `admin` permission when this field is not `0` (`TournamentRejectionReason.None`)
- [x] Set the tournament's processing status to done
- [x] Set all child match processing statuses as done
- [x] Reject all child matches

myssto:
Depends on #474
Closes #357


hburn7 (approved):
Confirmed works as expected

PR #476, opened 2024-10-09
Decouple `User` and `Player` DTOs

myssto:
There are many places in the API where we map `Player` entity properties directly on a DTO meant to represent a `User`. For simplicity and consistency we should never be doing this. It creates confusion between the two entities and what they functionally represent.

This PR serves as a proposal for an alternative which would be to nest a `PlayerDTO` (or in the case demonstrated here a `PlayerCompactDTO`) inside of any given `UserDTO` object so that relevant data like a `Player`'s id, username, osu! id, etc can be included without making it seem as if those fields are related directly to the `User`.

As an example, the `AdminNoteDTO` has been modified to use the described object structure instead of directly mapping the fields we believed to be relevant at the time. This also makes the DTO refactor proof if we ever needed to include more `User` or `Player` information in the future.

Admin note before:
Notice that `adminUsername` is directly mapped here when it shouldn't be and doesn't include the player id if someone wanted to navigate to the admin user's dashboard
```json
{
  "id": 1,
  "created": "2024-10-08T18:48:36.181316-05:00",
  "updated": null,
  "referenceId": 440,
  "adminUserId": 703,
  "adminUsername": "rimjob",
  "note": "this guy sucks!"
}
```

Admin note after:
```json
{
  "id": 1,
  "created": "2024-10-08T18:48:36.181316-05:00",
  "updated": null,
  "referenceId": 440,
  "adminUser": {
    "id": 703,
    "lastLogin": "2024-10-06T19:54:09.268338-05:00",
    "player": {
      "id": 7470,
      "osuId": 18068913,
      "username": "rimjob",
      "country": "US",
      "ruleset": 0,
      "userId": 703
    }
  },
  "note": "this guy sucks!"
}
```

As noted above there are quite a few spots where this happens for example `PlayerFrequencyDTO` and `PlayerFilteringResultDTO` which can be simplified in the same manner but can be addressed in future PRs as long as everyone agrees with the changes here.


myssto:
Blocking this because it's a relatively big breaking change for endpoint return types. Need to sort out with Aki first

hburn7 (approved):
I like this change. Tested and looks to be working, provided there are no system users in the database.

bugg86 (approved):

hburn7:
Depends on https://github.com/osu-tournament-rating/otr-web/issues/55

PR #481, opened 2024-10-09
Add `MatchWarningFlags` and `GameWarningFlags` bitwise enums

### LINKED ISSUE #478: Create warning flags bitwise enums
Implement the following for warning flags:

- [x] Match
    - Default value of 0 (codified as `None`) 
    - Title regex matching: {T_ABR}: ({TEAM_A}) vs ({TEAM_B}) where {T_ABR} is the tournament's abbreviation. vs check should be loose and cover VS vs., and everything in between (codified as `UnexpectedTitleFormat`).
    - Match game count is exactly 3 or 4 (codified as `LowGameCount`).
- [x] Game
    - Default value of 0 (codified as `None`) 
    - If a tournament does not have a stored mappool and the beatmap is used exactly once in the entire tournament (codified as `BeatmapUsedOnce`)
    - If a tournament does have a stored mappool, and the beatmap is not in the mappool, and the game is not one of the first two games in the match (codified as `BeatmapNotInMappool`)

- [x] Add `WarningFlags` to `Match` and `Game` Entities
- [x] Create and test migration

myssto:
Depends on #474
Closes #478

### DIFF
--- a/Database/Entities/Game.cs
+++ b/Database/Entities/Game.cs
@@ -69,6 +69,12 @@ public class Game : UpdateableEntityBase, IProcessableEntity, IAdminNotableEntit
     public GameRejectionReason RejectionReason { get; set; }
 
     /// <summary>
+    /// Warning flags
+    /// </summary>
+    [Column("warning_flags")]
+    public GameWarningFlags WarningFlags { get; set; }
+
+    /// <summary>
     /// Processing status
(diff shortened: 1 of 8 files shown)


hburn7 (approved):
Looks good

bugg86 (approved):

PR #485, opened 2024-10-12
Modify default rate limit windows

hburn7:
- Changes the default rate limit windows for all platforms to a time period of 60 seconds.
- Increases the default rate limit of the osu!track API (30 seconds per 2 minutes is very unrealistic)

Primarily making this change because prod has no way to override the frequency at which tokens reset for different platforms. This allows prod to call the osu!track API every 60 seconds which is necessary.


myssto (approved):
this is fine. this code existed before I added a way to configure the rate limit through appsettings so it's probably for the better that it go anyways 👍

Issue #489, opened 2024-10-12
Include admin notes on `TournamentDTO`


hburn7:
Include the tournament's admin notes on the `TournamentDTO`.

PR #477, opened 2024-10-09
Add `VerifiedByUser` and `SubmittedByUser` to tournament DTOs

myssto:
Depends on #476

Completes most of #336 save `Tournament-specific search`. Adds a `UserCompactDTO` for the submitter and verifier to tournament results. Also includes a couple query performance bumps in the tournaments repo.

PR #494, opened 2024-10-13
Create `TournamentCompactDTO` and include it in `MatchDTO`

### LINKED ISSUE #487: Add missing tournament information to `/matches/{id} GET`
Create a `TournamentCompactDTO` and add it to `MatchDTO`:
- Tournament name
- Tournament id
- Tournament lobby size
- Tournament abbreviation
- Tournament forum post url

hburn7:
Closes #487, #489

### DIFF
--- a/API/DTOs/TournamentCompactDTO.cs
+++ b/API/DTOs/TournamentCompactDTO.cs
@@ -1,0 +1,77 @@
+using System.Diagnostics.CodeAnalysis;
+using Database.Enums;
+using Database.Enums.Verification;
+
+namespace API.DTOs;
+
+public class TournamentCompactDTO
+{
+    /// <summary>
+    /// The tournament id
+    /// </summary>
+    public int Id { get; set; }
+
+    /// <summary>
+    /// The name of the tournament
+    /// </summary>
+    public string Name { get; set; } = null!;
+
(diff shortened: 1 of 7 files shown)


API/DTOs/TournamentCompactDTO.cs
```diff
@@ -0,0 +1,77 @@
...
+    /// </summary>
+    public int Id { get; init; }
+
+    /// <summary>
+    /// The name of the tournament
+    /// </summary>
+    public string Name { get; init; } = null!;
+
+    /// <summary>
+    /// Acronym / shortened name of the tournament
```
myssto, line 20:
```suggestion
    /// Acronym / shortened name
```
The docs for `TournamentDTO` are pretty old and we have shifted away from using phrases like `<description of prop> for the <entity name>`. If we could update the rest of these here that would be great

myssto (changes requested):
Just a couple docs changes. Keep in mind this PR will also conflict with #477 in the future because `TournamentCompactDTO` is already implemented in that.

hburn7:
> Just a couple docs changes. Keep in mind this PR will also conflict with #477 in the future because `TournamentCompactDTO` is already implemented in that.

Good to know

PR #502, opened 2024-10-20
Fix invalid entity field type & fix OAuth issues

hburn7:
- Fixes bug with refresh token encoding and decoding
  - A claim was encoded improperly and decoded improperly
- Adds support for viewing swagger while debugging with whitelist enabled
- Converts `ICollection<string>` field to `string[]` to fix SQL serialization bug.
  - Migrations are not needed


myssto (approved):
couldve used my branch but this works probably

PR #505, opened 2024-10-20
Add support for sorting tournaments by various keys

### LINKED ISSUE #503: Add support for sorting tournament data
On the `/tournaments` `GET` endpoint, add a way to sort by the following:
- StartDate
- Name

Implementation:
- [x] Create a new enum `TournamentSortKey` with the following values:
  - `StartDate`
  - `StartDateDescending`
  - `Name`
  - `NameDescending`
- [x] Add this to `TournamentRequestQueryDTO`
- [x] Add the behavior in the service / repo layer

hburn7:
Closes #503


myssto (changes requested):
There is already a similar impl of this with matches query sorting I believe. I'd like for them to behave similarly. Either adjust here to follow or vice versa

myssto (approved):
LGTM NOW thanks for syncing those up

Issue #482, opened 2024-10-09
Expand admin panel functionality to improve QOL


myssto:
Now that the base functionality of the admin panel is complete, we should start implementing some custom functions that will improve productivity and reduce the amount of clicks it will take to manually review a tournament submission. All routes detailed here should be implemented on the `EntityController.Admin.cs` partial class.

- Tournaments
  - [ ] Implement route `POST` `/tournaments/{id}:accept-verification-status`
    - Only run for tournaments with a `ProcessingStatus` of `NeedsVerification`
    - This endpoint should, for every entity in a tournament (including the tournament), change a `VerificationStatus` of `PreVerified` to `Verified` and `PreRejected` to `Rejected`. Functionally this can be seen as "accepting" the outcome of automation checks.
  - [ ] Implement route `POST` `/tournaments/{id}:rerun-auto-checks`
    - Can be ran at any `ProcessingStatus`
    - Take a query parameter `bool force = false`
    - This endpoint should, for every entity in a tournament (including the tournament):
      - If `force = false` skip any entities with a concrete verification (`Verified` or `Rejected`), else do not skip any
      - Set `VerificationStatus` to `None`
      - Set `RejectionReason` to `None`
      - Set `WarningFlags` to `None` (if applicable)
      - Set `ProcessingStatus` to `NeedsAutomationChecks`

(This is a draft, more should be added and then singular issues created for each endpoint once we determine the features we want)

myssto:
@cytusine0 @SourMongoose @hburn7 Feel free to edit or comment with additional thoughts

hburn7:
I think this flow is good. An admin can scan all of the pre-rejected / pre-verified entities, manually reject entities as needed, then click a button which verifies all pre-verified and rejects all pre-rejected. I considered having these be separate actions but I like the combined action better.

PR #506, opened 2024-10-20
Tournament deletion functionality

hburn7:
Part of #482 

- Adds functionality to delete a tournament via `/tournaments/{id}` `DELETE`.
- Endpoint is admin only


myssto (approved):
as long as you've tested that the cascade works as expected

PR #515, opened 2024-10-24
Refactor rate limit overrides

myssto:
Depends on #514

sns this one is kinda huge can't help it

- Removes the `RateLimitOverrides` object
- Removes the `RateLimitOverridesSerializer` class
- Removes `User.RateLimitOverrides`
- Refactor `OAuthClient.RateLimitOverride` to be of type `int?`

Rate limits (custom or not) will now always function as number of requests per minute. No need to set a custom window for anyone.


API.Tests/Utilities/ClaimsPrincipalExtensionsTests.cs
```diff
@@ -71,18 +70,17 @@ public void ClaimsPrincipal_GetTokenType()
     }
 
     [Fact]
-    public void ClaimsPrincipal_GetRateLimitOverrides()
+    public void ClaimsPrincipal_GetRateLimitOverride()
     {
+        const int expected = 100;
```
hburn7, line 75: Constants generally are formatted with PascalCase or SCREAMING_SNAKE_CASE. Your call though since it's a local variable.

API/Repositories/Interfaces/IOAuthClientRepository.cs
```diff
@@ -14,9 +14,9 @@ public interface IOAuthClientRepository : IRepository<OAuthClient>
     /// Updates the rate limit overrides of a client for the given id
     /// </summary>
     /// <param name="clientId">Id of the client</param>
-    /// <param name="rateLimitOverrides">Updated rate limit overrides</param>
+    /// <param name="rateLimitOverride"></param>
```
hburn7, line 17:
```suggestion
    /// <param name="rateLimitOverride">The new rate limit</param>
```

API/Utilities/Extensions/ClaimsPrincipalExtensions.cs
```diff
@@ -43,12 +42,16 @@ public static bool IsWhitelisted(this ClaimsPrincipal claimsPrincipal) =>
...
     /// <summary>
-    /// Gets the <see cref="RateLimitOverrides"/> of the principle if available
+    /// Gets the custom rate limit of the principle if available
     /// </summary>
-    public static RateLimitOverrides? GetRateLimitOverrides(this ClaimsPrincipal claimsPrincipal)
+    public static int? GetRateLimitOverride(this ClaimsPrincipal claimsPrincipal)
     {
-        var raw = claimsPrincipal.FindFirst(OtrClaims.RateLimitOverrides)?.Value;
-        return string.IsNullOrEmpty(raw) ? null : RateLimitOverridesSerializer.Deserialize(raw);
+        if (!int.TryParse(claimsPrincipal.FindFirst(OtrClaims.RateLimitOverrides)?.Value, out var rlo))
```
hburn7, line 49: I would prefer not to use acronyms like this in local variables. `out var override`, `limit`, `rateLimitOverride`, etc. are more suitable options. The exception being `Func`/`Expression` arguments (lambdas).

hburn7 (approved):

PR #518, opened 2024-10-26
Add endpoint for adding pooled beatmaps to a tournament

### LINKED ISSUE #510: Add endpoint for tournament beatmap id overwrite
An endpoint is needed to link beatmaps to tournaments. The endpoint should take a tournament id as a query parameter and a collection of beatmap ids (osu! ids) in the body.

- [x] Create endpoint `/api/v1/tournaments/{id}/beatmaps` `POST`
  - [x] Endpoint accepts `ICollection<long> beatmapIds` as the POST body
- [x] Pooled beatmaps table is updated with the appropriate mappings. A new repository may be needed.

hburn7:
Closes #510 (except it adds to the list instead of overwriting it)

- Adds `/tournaments/{id}/beatmaps` `POST` endpoint which accepts an `ICollection<long> osuBeatmapIds` collection.
- Each osu! id in the collection gets converted into a new beatmap if it doesn't exist.
- Each beatmap (either existing or newly created) is then added to the `Tournament`'s `PooledBeatmaps` collection. This creates an entry in the `__join__pooled_beatmaps` table automatically.
- The endpoint returns the current list of pooled beatmaps for the tournament after the modifications.
- This implementation ignores duplicates and does not unnecessarily create join table entries or beatmap table entries.

I am not concerned with the performance of this endpoint despite it requiring multiple round trips per query.


myssto:
Since I don't really have a better solution in mind off the top of my head I'm just gonna air this out and see what you think. First thing that comes to mind that bugs me is that we are returning the beatmap DTOs after creation. For any newly created beatmaps, these will be completely useless objects with empty fields for everything, and then it looks like we're also combining these with beatmaps that already exist.

So it feels a bit improper but what do you think? In my head there are a couple solutions I can think of right now:
- Map and return *only* the newly created beatmaps to a `CreatedDTO<T>` and return them with a 201 response instead of 200
- Return no response body (bleh)

Either way I think regardless of what we do we should better document what exactly this endpoint is returning since from an API consumer point of view its super super unclear

Database/Repositories/Implementations/TournamentsRepository.cs
```diff
@@ -90,6 +90,42 @@ public async Task<ICollection<Tournament>> GetAsync(int page, int pageSize, Tour
...
+        {
+            if (!tournament.PooledBeatmaps.Contains(b))
+            {
+                tournament.PooledBeatmaps.Add(b);
+            }
+        });
+
+        await UpdateAsync(tournament);
+        return tournament.PooledBeatmaps;
+    }
```
myssto, line 127: I count 4 db calls in this function alone. A lot of your logic with finding and creating unique beatmaps is really unnecessary [1]. In fact, I think this function should probably completely move to the service instead. I'd recommend looking at the way I tweaked your implementation of this on `TournamentsService.CreateAsync(TournamentSubmissionDTO submission, int submitterUserId, bool preApprove)`. You could probably abstract that out into a private function that can be used for both the existing function and this new one.

[1] This comes off a bit abrasive so I'll explain:
The way you find non-existing beatmap ids is fine, what I have an issue mostly with is the way you create the new beatmaps. There is no reason to create, add, and then save these new beatmaps on their own when you can create them by proxy. If you add the newly created beatmap objects to the `Tournament.PooledBeatmaps` collection, they will be saved to the database on the next save by proxy via EF :D This is exactly the same way we do it with matches and beatmaps in the service method

myssto (changes requested):
See my previous comment as well

hburn7:
I can't think of a use case where the beatmap created result would be used so I would prefer to just return 201.

Database/Repositories/Implementations/TournamentsRepository.cs
hburn7: I don't see a reason for this being moved to the service. This is change is suited more for the repository class in my eyes. I do agree that it's inefficient so I'll fix that.

myssto: I see where you're coming from but I'm just thinking in terms of consistency. The `TournamentsService.CreateAsync()` method does the exact same thing for creating beatmaps but on the service level instead. Do you think that method should be moved down to the repository as well?

hburn7: Opting to add this functionality to `TournamentsRepository` instead as it feels very odd to have to write a dedicated `GetAsync` function for including a specific navigation (`PooledBeatmaps` in this case).

myssto (approved):
Nice work! Assuming everything is tested it LGTM

PR #521, opened 2024-10-27
Introduce `/tournaments/{id}:accept-pre-statuses`

hburn7:
Part of #482 

- Introduces `/tournaments/{id}:accept-pre-statuses` endpoint which marks all `pre-` verification statuses as the `verified` equivalent.


API/Controllers/TournamentsController.Admin.cs
```diff
@@ -9,6 +9,29 @@ namespace API.Controllers;
...
+    /// Marks pre-rejected items as rejected, marks pre-verified
+    /// items as verified. Applies for the tournament and all children.
+    /// </summary>
+    /// <param name="id">Tournament id</param>
+    /// <response code="404">If a tournament matching the given id does not exist</response>
+    /// <response code="200">All items were updated successfully</response>
+    [HttpPost("{id:int}:accept-pre-statuses")]
+    [Authorize(Roles = OtrClaims.Roles.Admin)]
+    [ProducesResponseType(StatusCodes.Status404NotFound)]
+    [ProducesResponseType<AdminNoteDTO>(StatusCodes.Status200OK)]
```
myssto, line 22: bing bong nice copy paste

myssto (changes requested):
Nice work :D

I think it would be more productive if the endpoint returned the tournament since it's already in memory. This way frontend can update the status' of each entity immediately instead of requiring a refresh

myssto:
Also I'm not a fan of the name `accept-pre-statuses`. It doesn't directly reference a property like `VerificationStatus` like the proposed name does. I'll leave this one up to you

hburn7:
> Nice work :D
> 
> I think it would be more productive if the endpoint returned the tournament since it's already in memory. This way frontend can update the status' of each entity immediately instead of requiring a refresh

Note that this returns the tournament with all child navigations, but yeah I'm fine to do this until it presents a problem.

PR #522, opened 2024-10-27
Add `/tournaments/{id}:rerun-auto-checks` `POST`

### LINKED ISSUE #482: Expand admin panel functionality to improve QOL

hburn7:
Depends on #521
Part of #482 
Closes #482 

- Implements all of the behavior defined in #482 for the `/tournaments/{id}:rerun-auto-checks` endpoint

### DIFF
--- a/Database/Repositories/Implementations/TournamentsRepository.cs
+++ b/Database/Repositories/Implementations/TournamentsRepository.cs
@@ -90,6 +90,120 @@ public class TournamentsRepository(OtrContext context) : RepositoryBase<Tourname
                 .ToListAsync();
     }
 
+    public async Task AcceptVerificationStatuses(int id)
+    {
+        Tournament? tournament = await TournamentsBaseQuery()
+            .Where(t => t.ProcessingStatus == TournamentProcessingStatus.NeedsVerification)
+            .FirstOrDefaultAsync(t => t.Id == id);
+
+        if (tournament is null)
+        {
+            return;
+        }
+
+        tournament.VerificationStatus = EnumUtils.ConfirmPreStatus(tournament.VerificationStatus);
+        foreach (Match match in tournament.Matches)
+        {
+            match.VerificationStatus = EnumUtils.ConfirmPreStatus(match.VerificationStatus);
+
+            foreach (Game game in match.Games)
+            {
+                game.VerificationStatus = EnumUtils.ConfirmPreStatus(game.VerificationStatus);
+
+                foreach (GameScore score in game.Scores)
+                {
+                    score.VerificationStatus = EnumUtils.ConfirmPreStatus(score.VerificationStatus);
+                }
+            }
+        }
+
+        await UpdateAsync(tournament);
+    }
+
+    public async Task RerunAutomationChecks(int id, bool force = false)
+    {
+        Tournament? tournament = await TournamentsBaseQuery()
+            .FirstOrDefaultAsync(t => t.Id == id);
+
+        if (tournament is null)
+        {
+            return;
+        }
+
+        var update = force || (tournament.VerificationStatus != VerificationStatus.Rejected &&
+                               tournament.VerificationStatus != VerificationStatus.Verified);
+
+        if (update)
+        {
+            tournament.VerificationStatus = VerificationStatus.None;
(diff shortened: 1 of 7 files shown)


Database/Repositories/Implementations/TournamentsRepository.cs
```diff
@@ -185,6 +185,90 @@ public async Task DeletePooledBeatmapsAsync(int id, ICollection<int> beatmapIds)
...
+            {
+                continue;
+            }
+
+            score.VerificationStatus = VerificationStatus.None;
+            score.RejectionReason = ScoreRejectionReason.None;
+            score.ProcessingStatus = ScoreProcessingStatus.Done;
+        }
+    }
+
```
myssto, line 271: These should be either extension methods or defined on their appropriate repositories, not all on the tournaments repo

Database/Repositories/Interfaces/ITournamentsRepository.cs
```diff
@@ -68,6 +68,13 @@ Task<ICollection<Tournament>> GetAsync(int page, int pageSize, TournamentQuerySo
     /// <returns>The updated <see cref="Tournament"/></returns>
     Task<Tournament?> AcceptPreVerificationStatusesAsync(int id);
 
+    /// <summary>
+    /// Reruns automation checks for a tournament and all child entities
+    /// </summary>
+    /// <param name="id">Tournament id</param>
+    /// <param name="force">Whether to overwrite fully Verified/Rejected data (dangerous)</param>
+    Task RerunAutomationChecks(int id, bool force = false);
```
myssto, line 76: Method name and documentation is quite misleading imo. This method isn't actually running any automation checks, it is more appropriate to say that it is "signaling to the data worker that automation checks should be run again". Interpret that how you will but I think it should at the very least better describe what it is doing in documentation if not method name

myssto (changes requested):

Database/Repositories/Interfaces/ITournamentsRepository.cs
hburn7: agreed

myssto (approved):
Much better :D Nice work as always

myssto:
The same method on the entity approach should probably also be used for the "confirm pre status" method on the repository as well

hburn7:
Will take a look at that.

myssto (changes requested):
I would actually prefer that `ProcessableEntity` stays an interface to be implemented rather than an abstract base. The `ColumnAttribute`s for some reason do not like being inherited and end up dropping the value they are provided unless specified in the context. There is no case where we use the base implementation of `ProcessableEntityBase.ConfirmPreVerificationStatus()`, and `ProcessableEntityBase.ResetAutomationStatuses()` is an abstract. I feel it would make more sense to leave this as an interface to force the inheriting class to implement these.

myssto (approved):
Ok weird formatting aside this is great now!

PR #531, opened 2024-11-13
Separate JWT CLI role arguments by comma

hburn7:
- Fixes a bug where multiple roles could not be specified in the JWT CLI


myssto (approved):

myssto:
fwiw, this was working for me without having this separator... `--roles admin whitelist`

hburn7:
I don't think that was working for me but I would have to try again. That feels so jank anyway because CLI arguments are space separated, how does the CLI not throw a parse error over that? Windows thing maybe?

PR #532, opened 2024-11-13
Do not filter ruleset by default

hburn7:
- Fixes a bug where the default ruleset filter for tournaments is `Osu`.

### DIFF
--- a/API/DTOs/TournamentRequestQueryDTO.cs
+++ b/API/DTOs/TournamentRequestQueryDTO.cs
@@ -1,6 +1,5 @@
 using System.ComponentModel.DataAnnotations;
 using Database.Enums;
-using static Database.Enums.Ruleset;
 
 namespace API.DTOs;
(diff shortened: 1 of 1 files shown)


API/DTOs/TournamentRequestQueryDTO.cs
```diff
@@ -31,7 +31,7 @@ public class TournamentRequestQueryDTO
     /// <summary>
     /// An optional ruleset to filter by
     /// </summary>
-    public Ruleset? Ruleset { get; init; } = Osu;
+    public Ruleset? Ruleset { get; init; } = null;
```
myssto, line 34:
```suggestion
    public Ruleset? Ruleset { get; init; }
```

myssto (changes requested):
Lets remove that extra using as well

myssto (approved):

PR #533, opened 2024-11-13
Fix missing OsuTrackLastFetch update

hburn7:
- Fixes a bug where players with osu!track updates would not have an updated `OsuTrackLastFetch` field.

### DIFF
--- a/DataWorkerService/Services/Implementations/PlayersService.cs
+++ b/DataWorkerService/Services/Implementations/PlayersService.cs
@@ -145,7 +145,7 @@ public class PlayersService(
             await UpdateFromOsuTrackApiAsync(player);
         }
 
-        await context.SaveChangesAsync();
+        await playersRepository.UpdateAsync(outdatedPlayers);
 
         logger.LogDebug(
             "Updated Players with outdated osu! API data [Count: {Count}]",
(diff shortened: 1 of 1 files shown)


myssto (approved):

DataWorkerService/Services/Implementations/PlayersService.cs
```diff
@@ -145,7 +145,7 @@ public async Task UpdateOutdatedFromOsuTrackApiAsync(PlayerFetchPlatformConfigur
             await UpdateFromOsuTrackApiAsync(player);
         }
 
-        await context.SaveChangesAsync();
+        await playersRepository.UpdateAsync(outdatedPlayers);
```
myssto, line 148: Just thought id mention that the `repository.UpdateAsync()` technically doesn't do anything different than `context.SaveChangesAsync()` when the change tracker is enabled. In fact because of a quirk with `dbSet<T>.UpdateAsync()` it actually makes EF think that _all_ fields of an entity require updating and will execute a replace on all columns for that row. Weird stuff, but that is why in the data worker I preferred saving directly from the context

hburn7: I mainly called it this way to have consistent update behavior. It would feel odd to me to not call UpdateAsync because then not all of the entity's update behavior is happening inside the repository methods. Makes for some head scratching later on if we need to debug why something isn't updating as expected.

PR #543, opened 2024-11-19
Standardize API route documentation

### LINKED ISSUE #524: Use `RequiredAttribute` for non-optional query parameters
This one is strictly code quality and is useful for the swagger generator. By default the swagger generator will assume that query parameters are optional when in some cases they aren't. We should annotate required query parameters with the `System.ComponentModel.DataAnnotations.RequiredAttribute` to be more verbose.

For example currently (From the `OAuthController`):
```cs
public async Task<IActionResult> AuthorizeClientAsync(
    [FromQuery] int clientId, 
    [FromQuery] string clientSecret
)
```

Since the `clientId` and `clientSecret` are required query parameters, it could be annotated like so:
```cs
public async Task<IActionResult> AuthorizeClientAsync(
    [FromQuery] [Required] int clientId,
    [FromQuery] [Required] string clientSecret
)
```

myssto:
Depends on #542
Closes #524

Standardizes the wording, verbiage, and attribute usage across all API routes. I'd be happy to write up a docs page about standards regarding this in the future :D


API/Controllers/TournamentsController.cs
```diff
@@ -20,37 +22,27 @@ IAdminNoteService adminNoteService
     /// <summary>
     /// Get all tournaments which fit an optional request query
     /// </summary>
-    /// <param name="requestQuery">The optional request query filter</param>
     /// <remarks>Will not include match data</remarks>
     /// <response code="200">Returns all tournaments which fit the request query</response>
-    /// <returns>
-    /// A page of tournaments
-    /// </returns>
```
hburn7, line 28: Why remove these?

API/Controllers/TournamentsController.cs
```diff
@@ -20,37 +22,27 @@ IAdminNoteService adminNoteService
...
     [ProducesResponseType<IEnumerable<TournamentDTO>>(StatusCodes.Status200OK)]
-    [ProducesResponseType(StatusCodes.Status401Unauthorized)]
-    public async Task<IActionResult> ListAsync([FromQuery] TournamentRequestQueryDTO? requestQuery)
-    {
-        requestQuery ??= new TournamentRequestQueryDTO();
-
-        ICollection<TournamentDTO> tournaments = await tournamentsService.GetAsync(requestQuery);
-        return Ok(tournaments);
-    }
+    public async Task<IActionResult> ListAsync([FromQuery] TournamentRequestQueryDTO requestQuery) =>
```
hburn7, line 29: Why isn't `requestQuery` nullable?

API/Controllers/TournamentsController.cs
```diff
@@ -60,7 +52,7 @@ public async Task<IActionResult> CreateAsync([FromBody] TournamentSubmissionDTO
         if (await tournamentsService.ExistsAsync(tournamentSubmission.Name, tournamentSubmission.Ruleset))
         {
             return BadRequest(
-                $"A tournament with name {tournamentSubmission.Name} for ruleset {tournamentSubmission.Ruleset} already exists"
+                $"A tournament with name: {tournamentSubmission.Name} for ruleset: {tournamentSubmission.Ruleset} already exists"
```
hburn7, line 55:
```suggestion
                $"A tournament with name '{tournamentSubmission.Name}' for ruleset '{tournamentSubmission.Ruleset}' already exists"
```

This seems like it would read nicer in a log. Colons don't work here.

API/Controllers/TournamentsController.cs
```diff
@@ -69,15 +61,13 @@ public async Task<IActionResult> CreateAsync([FromBody] TournamentSubmissionDTO
...
 
-        // Remove query string
-        try
-        {
-            tournamentSubmission.ForumUrl = new Uri(tournamentSubmission.ForumUrl).GetLeftPart(UriPartial.Path);
-        }
-        catch (UriFormatException)
+        // Sanitize forum post url
+        Uri.TryCreate(tournamentSubmission.ForumUrl, UriKind.Absolute, out Uri? sanitizedUrl);
+        if (sanitizedUrl is null)
```
hburn7, line 66: Would prefer something like `if (!Uri.TryCreate ...`

API/Controllers/UsersController.cs
```diff
@@ -141,15 +142,14 @@ public async Task<IActionResult> GetClientsAsync(int id)
...
-    /// <param name="id">Id of the user</param>
-    /// <param name="clientId">Id of the OAuth client</param>
-    /// <response code="404">If a user or client does not exist</response>
-    /// <response code="400">If the deletion was not successful</response>
-    /// <response code="200">If the deletion was successful</response>
+    /// <param name="id">User id</param>
+    /// <param name="clientId">OAuth client id</param>
+    /// <response code="404">
+    /// A user matching the given id does not exist
+    /// or an oauth client matching the given id is not owned by the user
```
hburn7, line 149:
```suggestion
    /// or an OAuth client matching the given id is not owned by the user
```

API/Controllers/UsersController.cs
```diff
@@ -170,19 +170,16 @@ public async Task<IActionResult> DeleteClientAsync(int id, int clientId)
...
-    /// <param name="id">Id of the user</param>
-    /// <param name="clientId">Id of the OAuth client</param>
-    /// <response code="404">If a user or client does not exist</response>
-    /// <response code="400">If the secret reset was not successful</response>
-    /// <response code="200">Returns new client credentials if the secret reset was successful</response>
+    /// <param name="id">User id</param>
+    /// <param name="clientId">OAuth client id</param>
+    /// <response code="404">
+    /// A user matching the given id does not exist
+    /// or an oauth client matching the given id is not owned by the user
```
hburn7, line 177:
```suggestion
    /// or an OAuth client matching the given id is not owned by the user
```

API/Controllers/MatchesController.Admin.cs
```diff
@@ -35,42 +37,21 @@ public async Task<IActionResult> CreateAdminNoteAsync(int id, [FromBody] string
...
+    /// Update an admin note for a match
     /// </summary>
     /// <param name="id">Match id</param>
     /// <param name="noteId">Admin note id</param>
+    /// <param name="note">New content of the admin note</param>
     /// <response code="404">
-    /// If a match matching the given id does not exist.
-    /// If an admin note matching the given noteId does not exist
+    /// A match matching the given id does not exist
+    /// or an admin note matching the given noteId does not exist
```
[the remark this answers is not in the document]
myssto: If you remember me mentioning XML tag sanitization, this follows the same principle. You'd think adding one would make sense but it actually renders to the same thing :P

API/Controllers/MatchesController.cs
```diff
@@ -17,43 +16,21 @@ namespace API.Controllers;
...
-        // Clamp page size for non-admin users
-        if (limit > 5000 && !User.IsAdmin())
-        {
-            limit = 5000;
-        }
-
-        return Ok(await matchesService.GetAsync(limit, page, filter));
-    }
+    [ProducesResponseType<IEnumerable<MatchDTO>>(StatusCodes.Status200OK)]
+    public async Task<IActionResult> ListAsync([FromQuery] MatchRequestQueryDTO requestQuery) =>
```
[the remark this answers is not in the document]
myssto: If you mean the `MatchRequestQueryDTO requestQuery`, then actually no. For complex objects, the model binder doesn't really gaf about the whole object per se. Whether or not a property is required in a complex object is determined by it's definition on that type.

TLDR it's really weird because it varies per use case. If you want me to explain at length I will in a call :P

API/Controllers/OAuthController.cs
```diff
@@ -78,18 +82,15 @@ public async Task<IActionResult> CreateClientAsync() =>
     /// <summary>
     /// Generate new access credentials from a valid refresh token
     /// </summary>
-    /// <remarks>
-    /// Generated access credentials will contain only a new access token,
-    /// and the given refresh token is returned with it
-    /// </remarks>
-    /// <param name="refreshToken"></param>
-    /// <response code="400">If the given refresh token is invalid, or there was an error during authorization</response>
+    /// <remarks>Generated access credentials will contain only a new access token</remarks>
```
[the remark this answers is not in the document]
myssto: Doesn't work like this anymore! The OAuth handler [only returns the access token now](https://github.com/osu-tournament-rating/otr-api/blob/cf700bf1565c40ca6854829223db435381109bee/API/Handlers/Implementations/OAuthHandler.cs#L141-L154). Returning the old refresh token really doesn't make any sense and from user perspective kind of feels like "hey I think this refreshed my refresh token as well?".

API/Controllers/PlayersController.cs
```diff
@@ -2,31 +2,99 @@
...
+            ? Ok(result)
+            : NotFound();
+    }
+
+    /// <summary>
+    /// Get a player's stats
+    /// </summary>
+    /// <remarks>
+    /// Gets player by versatile search.
+    /// If no ruleset is provided, the player's default is used. <see cref="Endpoints.Osu"/> is used as a fallback.
```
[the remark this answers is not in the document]
myssto: Yea dk why this happened

API/Controllers/StatsController.cs
```diff
@@ -1,63 +0,0 @@
...
-        return result is null
-            ? NotFound()
-            : Ok(result);
-    }
-
-    [HttpGet("histogram")]
-    [Authorize(Roles = $"{OtrClaims.Roles.User}, {OtrClaims.Roles.Client}")]
-    public async Task<ActionResult<IDictionary<int, int>>> GetRatingHistogramAsync(
-        [FromQuery] Ruleset ruleset = Ruleset.Osu
-    ) => Ok(await baseStatsService.GetHistogramAsync(ruleset));
```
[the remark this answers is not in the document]
myssto: Found no usage of it on web so I axed it but kept the underlying hierarchy intact

API/Controllers/TournamentsController.cs
myssto: When binding a complex object `[FromQuery]` adding a `<param>` tag does nothing, all of the documentation for the parameters live in the type itself. We shouldn't be using `<returns>` tags, as that is functionally what `<response code="200">` (or whatever successful code) is documenting

API/Controllers/TournamentsController.cs
myssto: See https://github.com/osu-tournament-rating/otr-api/pull/543#discussion_r1849096443

API/Controllers/MatchesController.Admin.cs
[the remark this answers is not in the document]
hburn7: From a readability standpoint I also don't love a billion <see cref>'s in there anyway.

API/Controllers/MatchesController.cs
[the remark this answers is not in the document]
hburn7: Can you document the cases that should and should not be nullable?

API/Controllers/OAuthController.cs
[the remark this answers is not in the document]
hburn7: Is there a defined standard for this or is this just a "what we feel is best" thing?

API/Controllers/StatsController.cs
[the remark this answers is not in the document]
hburn7: It will be used eventually, its on the figma. May be good to re-add.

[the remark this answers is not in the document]
myssto: Ah I see, its part of the site statistics page. In that case, I think this shouldnt be its own endpoint anyways. I think we should probably bundle all of that information into one large object (kind of like the player stats dto for the players page) for easy caching instead if breaking it out into a ton of different endpoints. That being said I think it's fine that this be removed for now.

hburn7 (approved):

API/Controllers/TournamentsController.cs
```diff
@@ -20,68 +22,89 @@ IAdminNoteService adminNoteService
...
-                $"A tournament with name {tournamentSubmission.Name} for ruleset {tournamentSubmission.Ruleset} already exists"
+            ModelState.AddModelError(
+                nameof(TournamentSubmissionDTO.Name),
+                $"A tournament with the name '{tournamentSubmission.Name}' and ruleset " +
+                $"'{tournamentSubmission.Ruleset}' already exists."
+            );
+            ModelState.AddModelError(
+                nameof(TournamentSubmissionDTO.Ruleset),
+                $"A tournament with the ruleset '{tournamentSubmission.Ruleset}' and name " +
+                $"'{tournamentSubmission.Name}' already exists."
```
hburn7, line 64: I think it would be good to explain in a comment why it's necessary to add two separate `ModelState` errors because it looks redundant at a glance.

API/Controllers/TournamentsController.cs
```diff
@@ -20,68 +22,89 @@ IAdminNoteService adminNoteService
...
+            // Check reject on submit
+            if (tournamentSubmission.RejectionReason.HasValue)
+            {
+                ModelState.AddModelError(
+                    nameof(TournamentSubmissionDTO.RejectionReason),
+                    $"The field {nameof(TournamentSubmissionDTO.RejectionReason)} may only be supplied by admin users."
+                );
+            }
+
+            // We never want unprivileged users submitting empty tournaments
```
hburn7, line 80: This implies admins can submit tournaments which have no matches. Probably shouldn't be possible regardless of permission level.

API/Controllers/TournamentsController.cs
```diff
@@ -20,68 +22,89 @@ IAdminNoteService adminNoteService
...
+            return ValidationProblem();
         }
 
-        // Input validation for mp and beatmap ids
-        tournamentSubmission.Ids = tournamentSubmission.Ids.Where(id => id >= 1);
-        tournamentSubmission.BeatmapIds = tournamentSubmission.BeatmapIds.Where(id => id >= 1);
+        #endregion
+
+        // Sanitize forum url
+        tournamentSubmission.ForumUrl = sanitizedUri!.AbsolutePath;
```
hburn7, line 107: Why does `Uri.TryCreate` produce a nullable `Uri` at all? I feel like it shouldn't be assumed the Uri is non-null here if the `Uri.TryCreate` method produces a nullable `out`.

API/DTOs/TournamentSubmissionDTO.cs
```diff
@@ -14,18 +14,21 @@ public class TournamentSubmissionDTO
     /// <summary>
     /// The name of the tournament
     /// </summary>
-    public required string Name { get; set; }
```
hburn7, line 17: Is this a formatting rule? Does the `required` keyword function the same as `[Required]`? I wouldn't mind using the keyword but that can come later.

hburn7 (changes requested):
Just a few more small things/questions. Good stuff!

API/Controllers/TournamentsController.cs
myssto: Its not meant to be an implication, that is the intended behavior. There is a specific tournament rejection reason that exists for cases when the match data is lost / unrecoverable. In this case the tournament can be submitted this way with an empty list of mp ids so that the name and ruleset combination cannot be submitted again

API/Controllers/TournamentsController.cs
myssto: The `ModelState` property is of the type `ModelStateDictionary` which, as it implies, is literally just a dictionary of keys to values. Keys being the names of model properties and values being any number of error messages. Since this case shows that both items in the compound unique index are invalid, I think they should both receive a model state error. Ill add a short comment too

API/Controllers/TournamentsController.cs
myssto: This is the definition for the method

```cs
public static bool TryCreate([NotNullWhen(true), StringSyntax(StringSyntaxAttribute.Uri, "uriKind")] string? uriString, UriKind uriKind, [NotNullWhen(true)] out Uri? result)
```

99% of the time the case with `Try{X}` methods that return a `bool` and produce an `out` parameter is that the `out` parameter _has to be nullable_. You can work with this and improve code quality by using a `System.Diagnostics.CodeAnalysis.NotNullWhenAttribute` which can flag to analyzers that the `out` value will always be present based on the return value of the method.

Since we do this:

```cs
        // Check for valid forum post url
        if (!Uri.TryCreate(tournamentSubmission.ForumUrl, UriKind.Absolute, out Uri? sanitizedUri))
        {
            ModelState.AddModelError(
                nameof(TournamentSubmissionDTO.ForumUrl),
                $"The field {nameof(TournamentSubmissionDTO.ForumUrl)} is invalid."
            );
        }

        if (!ModelState.IsValid)
        {
            return ValidationProblem();
        }

        // Sanitize forum url
        tournamentSubmission.ForumUrl = sanitizedUri!.AbsolutePath;
```
It is not clear to the analyzer whether or not the result of the method caused a return or branching flow. However we do know for sure that if the result is false, the statement `tournamentSubmission.ForumUrl = sanitizedUri!.AbsolutePath` would never execute because of the check to the model state.

myssto: I actually implemented a method with the same design as what I am describing in my last batch of changes
```cs
    public static bool TryGetSubjectId(
        this ClaimsPrincipal claimsPrincipal,
        [NotNullWhen(true)] out int? subjectId
    )
    {
        subjectId = null;
        try
        {
            subjectId = claimsPrincipal.GetSubjectId();
            return true;
        }
        catch (ArgumentException)
        {
            return false;
        }
    }
```

API/DTOs/TournamentSubmissionDTO.cs
myssto: I had added the `required` keyword quite awhile ago because I was under the assumption that it accomplished the same thing as the `[Required]` attribute but that is not the case. Well, sort of. From my understanding the functional difference for API usage is that using the attribute causes model validation to fail when given an empty string for a value where the keyword does not.

API/Controllers/TournamentsController.cs
hburn7: Wow, very interesting. Thanks for the explanation.

API/DTOs/TournamentSubmissionDTO.cs
hburn7: good to know

API/Controllers/TournamentsController.cs
hburn7: Huh okay, yeah guess that makes sense.

hburn7 (approved):

PR #548, opened 2024-11-22
Improve audits

### LINKED ISSUE #546: Investigate excess audits
When I last checked the database, I think we had around ~60M audits for Games - some really big number.

Need to investigate how audits are being generated and verify that it's behaving as expected. The database continues to fill up to capacity and will undoubtedly lead to a critical failure in prod if we fail to properly address runaway audits.

myssto:
Closes #546

Improve the way audits are generated by adding an `AuditIgnoreAttribute` that can be used to annotate properties of the auditable entity for which audits should not track changes to. As an example, in this PR all processable entities annotate the `LastProcessingDate` with the attribute, which prevents audits from creating changelog entries when the property is modified.


Database/Entities/GameScore.cs
```diff
@@ -110,6 +107,18 @@ public class GameScore : UpdateableEntityBase, IProcessableEntity, IAdminNotable
...
+    /// </summary>
+    [Column("game_id")]
+    public int GameId { get; set; }
+
+    /// <summary>
+    /// Id of the <see cref="Entities.Player" /> that set the <see cref="GameScore" />
+    /// </summary>
+    [Column("player_id")]
+    public int PlayerId { get; set; }
+
```
hburn7, line 121: Why move these down?

hburn7 (approved):

Database/Entities/GameScore.cs
myssto: I meant to actually put them above their navigations lol

PR #554, opened 2024-12-06
Use split query for `GetNeedingProcessingAsync`

hburn7:
- Fixes a critical performance issue with `TournamentsRepository.GetNeedingProcessingAsync()`. Single queries must not be used at this level of complexity - the queries are simply too large and will fail to complete.


hburn7:
idk why tests are failing

myssto (approved):
Again I'm completely fine with this as long as you are reasonably confident that there are no quirky EF side effects that come with it. I wish I could remember why I had to do this in the first place...

PR #530, opened 2024-11-11
Add support for friends list storage on login

hburn7:
# Friend list support

Closes osu-tournament-rating/otr-web#521 

Not required for beta but a nice feature to have - friends list support (from an architecture standpoint).

- Adds support for storage of friends list on login (specifically, whenever a user authorizes with an osu! code). This can be triggered by signing out and signing in on the website or when a user's refresh token expires.
- Updates friend list every 24 hours at most.
- Does not overwrite friend list if the osu! API returns an error, in other cases it does overwrite the list.
- Friends are stored in a join table called `__join__friends` - nearly identical to `__join__pooled_beatmaps`
- Adds support for calling osu!'s friends API.

- [x] Update staging & prod to use new configuration value

Issue #135, opened 2024-02-28
Support showing friends list


hburn7:
We want to allow users to view their friends on o!TR.

Requirements:

- [ ] Friends lookup table, data access, etc.
   - Name: `friends`
   - Schema:
      - `id` primary key, int, auto inc
      - `player_id`
      - `friend_id` (FK ref to `player_id` field of other friend on o!TR, assuming the player has any friends in our system.
      - `timestamp`
- [ ] When authorizing via osu!, sync the friends list.
- [ ] Allow users to manually sync their friends via `/players/{playerId}/friends/refresh POST` (@AkinariHex This is something that would be done through a user settings page.)
- [ ] Allow fetching of friends list via `/players/{playerId}/friends/refresh GET`
- [ ] Add `isFriend: boolean` to `/leaderboards GET` (assuming the user is authorized).

majorro:
I'll do the 4th task

PR #560, opened 2024-12-12
Add get friend list endpoint

majorro:
Part of #135 

Complementing #530 adds an endpoint for fetching all player's friends

### DIFF
--- a/API/Controllers/PlayersController.cs
+++ b/API/Controllers/PlayersController.cs
@@ -96,4 +96,22 @@ public partial class PlayersController(
 
         return Ok(await adminNoteService.ListAsync<PlayerAdminNote>(id));
     }
+
+    [HttpGet("{id:int}/friends")]
+    [Authorize(Roles = $"{OtrClaims.Roles.User}, {OtrClaims.Roles.Client}")]
+    [ProducesResponseType(StatusCodes.Status404NotFound)]
+    [ProducesResponseType<IEnumerable<PlayerCompactDTO>>(StatusCodes.Status200OK)]
+    public async Task<IActionResult> ListFriendsAsync(int id)
+    {
+        if (!await playerService.ExistsAsync(id))
+        {
+            return NotFound();
+        }
+
+        IEnumerable<PlayerCompactDTO>? result = await playerService.GetFriendsAsync(id);
+
+        return result is null
+            ? NotFound()
+            : Ok(result);
+    }
 }
--- a/API/Controllers/PlayersController.cs
+++ b/API/Controllers/PlayersController.cs
@@ -96,22 +96,4 @@ public partial class PlayersController(
 
         return Ok(await adminNoteService.ListAsync<PlayerAdminNote>(id));
     }
-
-    [HttpGet("{id:int}/friends")]
-    [Authorize(Roles = $"{OtrClaims.Roles.User}, {OtrClaims.Roles.Client}")]
-    [ProducesResponseType(StatusCodes.Status404NotFound)]
-    [ProducesResponseType<IEnumerable<PlayerCompactDTO>>(StatusCodes.Status200OK)]
-    public async Task<IActionResult> ListFriendsAsync(int id)
-    {
-        if (!await playerService.ExistsAsync(id))
-        {
-            return NotFound();
-        }
-
-        IEnumerable<PlayerCompactDTO>? result = await playerService.GetFriendsAsync(id);
-
-        return result is null
-            ? NotFound()
-            : Ok(result);
-    }
 }
--- a/API/Controllers/UsersController.cs
+++ b/API/Controllers/UsersController.cs
@@ -257,8 +257,11 @@ public class UsersController(
     /// Get a user's friends
     /// </summary>
     /// <param name="id">User id</param>
+    /// <response code="404">A user matching the given id does not exist</response>
+    /// <response code="200">The operation was successful</response>
     [HttpGet("{id:int}/friends")]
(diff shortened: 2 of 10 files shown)


majorro:
Pointing this to [feature/friends-list-storing](https://github.com/osu-tournament-rating/otr-api/tree/feature/friends-list-storing) becuase this pr would get flooded with that branch commits

API/Controllers/PlayersController.cs
```diff
@@ -96,4 +96,22 @@ public async Task<IActionResult> ListAdminNotesAsync(int id)
 
         return Ok(await adminNoteService.ListAsync<PlayerAdminNote>(id));
     }
+
+    [HttpGet("{id:int}/friends")]
```
hburn7, line 100: Needs endpoint documentation, please see the other endpoints for examples.

API/Controllers/PlayersController.cs
```diff
@@ -96,4 +96,22 @@ public async Task<IActionResult> ListAdminNotesAsync(int id)
...
+        {
+            return NotFound();
+        }
+
+        IEnumerable<PlayerCompactDTO>? result = await playerService.GetFriendsAsync(id);
+
+        return result is null
+            ? NotFound()
+            : Ok(result);
+    }
```
hburn7, line 116: I think this needs to live under the `UsersController` actually. Friends are only populated when a `User` interacts with the website. Thoughts? @myssto

API/Services/Implementations/PlayerService.cs
```diff
@@ -6,7 +6,7 @@
 
 namespace API.Services.Implementations;
 
-public class PlayerService(IPlayersRepository playerRepository, IMapper mapper) : IPlayerService
+public class PlayerService(IPlayersRepository playerRepository, IUserService userService, IMapper mapper) : IPlayerService
```
hburn7, line 9: Assuming the endpoint is supposed to be on the `UsersController`, the changes in this file would be discarded.

hburn7 (changes requested):
My thoughts so far. I think this will require a lot of changes if this is to be a user-derived feature rather than a player-derived one.

API/Controllers/PlayersController.cs
majorro: I relied on the issue description, so I've done it this way. Imo it should be sticked to users too

hburn7: Yeah that issue is quite old, sorry.

API/Controllers/PlayersController.cs
myssto: +1 Makes me wish I would stop putting off writing that docs page

API/Controllers/PlayersController.cs
myssto: Yes, this endpoint should be on the `UsersController` instead

API/Controllers/PlayersController.cs
```diff
@@ -96,4 +96,22 @@ public async Task<IActionResult> ListAdminNotesAsync(int id)
 
         return Ok(await adminNoteService.ListAsync<PlayerAdminNote>(id));
     }
+
+    [HttpGet("{id:int}/friends")]
+    [Authorize(Roles = $"{OtrClaims.Roles.User}, {OtrClaims.Roles.Client}")]
```
myssto, line 101:
```suggestion
    [Authorize(Policy = AuthorizationPolicies.AccessUserResources]
```
I think to keep the expectation of privacy the same across data sources this needs to be only readable by the resource owner. You are not implicitly able to read the friends lists of others on osu! which is where this data is coming from in the first place.

myssto (changes requested):

API/Controllers/PlayersController.cs
majorro: should've added a todo xd

API/Controllers/PlayersController.cs
majorro: moved

API/Services/Interfaces/IUserService.cs
```diff
@@ -33,6 +33,13 @@ public interface IUserService
     /// <returns>A list of match submissions, or null if not found</returns>
     Task<IEnumerable<MatchSubmissionStatusDTO>?> GetSubmissionsAsync(int id);
 
+    /// <summary>
+    /// Gets a user's friends for the given id
+    /// </summary>
+    /// <param name="id">Id of the user</param>
+    /// <returns>A list of user's friends</returns>
```
hburn7, line 40:
```suggestion
    /// <returns>A list of the user's friends</returns>
```

hburn7 (changes requested):
A lot better, just some minor things.

API/Controllers/UsersController.cs
```diff
@@ -252,4 +252,23 @@ public async Task<IActionResult> SyncRulesetAsync(int id)
...
             : BadRequest();
     }
+
+    /// <summary>
+    /// Get a user's friends
+    /// </summary>
+    /// <param name="id">User id</param>
+    [HttpGet("{id:int}/friends")]
+    [Authorize(Policy = AuthorizationPolicies.AccessUserResources)]
+    [ProducesResponseType<IEnumerable<PlayerCompactDTO>>(StatusCodes.Status200OK)]
```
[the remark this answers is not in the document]
majorro: I just want to ask myself

How in the world am I missed this???

Will fix later

hburn7:
Going to go ahead and merge this into the base friends list branch but will be parking that feature to focus on more important items. Thanks for working on this one @majorro :)

PR #561, opened 2024-12-13
Fix rejection reason generation for tournaments with pooled beatmaps

hburn7:
- Fixes a bug where a random rejection reason would sometimes cause  `GameBeatmapUsageCheckTests.Check_GameNotRejected_WhenBeatmapPooled` to fail


myssto (approved):
lgtm

PR #565, opened 2024-12-19
DataWorkerService performance improvements

hburn7:
- Adds use of batch task scheduling where possible
- Removes numerous `SaveChangesAsync` calls (instead, one call is made at the end of each tournament batch)
- Adds `GetNextProcessor` functionality (enables better parallelism by not iterating through all processors for a tournament)

Has been tested end to end, from zero automation checks complete to entities being marked as verified/rejected.


DataWorkerService/BackgroundServices/TournamentProcessorService.cs
```diff
@@ -49,45 +49,51 @@ protected override async Task DoWork(IServiceScope scope, CancellationToken stop
...
+
+        await Task.WhenAll(tasks);
+
+        await context.SaveChangesAsync(stoppingToken);
+    }
+
+    private async Task ProcessAsync(CancellationToken stoppingToken, Tournament tournament,
+        ITournamentProcessorResolver tournamentProcessorResolver)
+    {
+        var sw = new Stopwatch();
```
myssto, line 67: Why create a new one when `_stopwatch` exists? This is seemingly not read from ever?

DataWorkerService/Processors/Resolvers/Implementations/TournamentProcessorResolver.cs
```diff
@@ -10,17 +11,29 @@ IEnumerable<IProcessor<Tournament>> processors
...
         ?? throw new InvalidOperationException($"Processor was not registered: {nameof(TournamentStatsProcessor)}");
 
+    public IProcessor<Tournament> GetNextProcessor(TournamentProcessingStatus processingStatus) =>
+        processingStatus switch
+        {
+            TournamentProcessingStatus.NeedsMatchData => GetDataProcessor(),
+            TournamentProcessingStatus.NeedsAutomationChecks => GetAutomationChecksProcessor(),
+            TournamentProcessingStatus.NeedsStatCalculation => GetStatsProcessor(),
+            TournamentProcessingStatus.NeedsVerification => GetVerificationProcessor(),
+            _ => throw new Exception("No processor")
```
myssto, line 32: Error message should be more descriptive. Maybe use `InvalidArgumentException` as well?

DataWorkerService/Processors/Resolvers/Interfaces/ITournamentProcessorResolver.cs
```diff
@@ -14,4 +15,6 @@ public interface ITournamentProcessorResolver : IProcessorResolver<Tournament>
     /// Gets the <see cref="IProcessor{TEntity}"/> implementation of the Tournament Stats Processor
     /// </summary>
     IProcessor<Tournament> GetStatsProcessor();
+
+    IProcessor<Tournament> GetNextProcessor(TournamentProcessingStatus processingStatus);
```
myssto, line 19: Document this

DataWorkerService/BackgroundServices/TournamentProcessorService.cs
```diff
@@ -49,45 +49,51 @@ protected override async Task DoWork(IServiceScope scope, CancellationToken stop
...
+        if (tournament.Matches.Count == 0)
+        {
+            tournament.VerificationStatus = VerificationStatus.Rejected;
+            tournament.RejectionReason = TournamentRejectionReason.IncompleteData;
+            tournament.ProcessingStatus = TournamentProcessingStatus.Done;
+            tournament.LastProcessingDate = DateTime.Now;
+
+            logger.LogWarning("Skipping processing for tournament with no matches [Id: {Id}]", tournament.Id);
+            return;
         }
```
myssto, line 86: This can probably be called outside of this function (before the task list) instead of during each iteration

myssto (changes requested):
These, and remove the context from the processors where `context.SaveChangesAsync()` were removed

DataWorkerService/BackgroundServices/TournamentProcessorService.cs
hburn7: Was my attempt at making it thread-safe, will need to think of something else I think.

hburn7: Actually in my head I can't see why a new stopwatch in each thread wouldn't work. But definitely can't use a shared one anymore.

hburn7: Opting to use a new format for this which I think looks much better.

CleanShot 2024-12-19 at 15 12 58@2x

myssto: Wouldn't it be easy to check for a 'no change' scenario on completion? Would be nice if there was a noticeable visual distinction between changing status and no change

hburn7: Yeah good idea

DataWorkerService/BackgroundServices/TournamentProcessorService.cs
```diff
@@ -48,46 +51,72 @@ protected override async Task DoWork(IServiceScope scope, CancellationToken stop
...
+            );
+        }
+        else
+        {
+            logger.LogDebug(
+                "Processing completed [Id: {Id} | Processing Status: {Before} --> {After}]",
+                tournament.Id,
+                processingStatusBefore,
+                tournament.ProcessingStatus
             );
```
myssto, line 119: Not really what I meant LOL... it's the same log statement with the level changed

myssto:
```cs
            logger.LogDebug(
                "Processing completed [Id: {Id} | Processing Status: {Before} (No change)]",
                tournament.Id,
                processingStatusBefore
            );
```

myssto (approved):

DataWorkerService/BackgroundServices/TournamentProcessorService.cs
hburn7: Fixed

PR #568, opened 2024-12-20
Fix stats & add missing ruleset column to `rating_adjustments` table

hburn7:
- Adds a `ruleset` column to the `rating_adjustments` table. This is something we missed when first implementing and is required. Without it, it's not possible to know which player's ruleset-specific rating is impacted by a non-match adjustment.
- Reintroduces support for player statistics via `/player/{id}/stats`. This functionality broke during our big rewrite. Although further rewrites are definitely required, this is better than nothing. Web should be updated to point to the data which is returned by this endpoint.

### DIFF
--- a/API.Tests/Services/LeaderboardServiceTests.cs
+++ b/API.Tests/Services/LeaderboardServiceTests.cs
@@ -3,7 +3,9 @@ using API.Enums;
 using API.Services.Implementations;
 using API.Utilities;
 using APITests.MockRepositories;
+using AutoMapper;
 using Database.Enums;
+using MapperProfile = OsuApiClient.Configurations.MapperProfile;
 
 namespace APITests.Services;
 
@@ -44,7 +46,8 @@ public class LeaderboardServiceTests
             playerRatingsRepository.Object,
             matchStatsRepository.Object,
             playerRepository.Object,
-            tournamentsService
+            tournamentsService,
+            new Mapper(new MapperConfiguration(configuration => configuration.AddProfile(new MapperProfile())))
         );
 
         _leaderboardService = new LeaderboardService(
--- a/API/DTOs/PlayerRatingChartDataPointDTO.cs
+++ b/API/DTOs/PlayerRatingChartDataPointDTO.cs
@@ -1,3 +1,5 @@
+using Database.Enums;
+
 namespace API.DTOs;
 
 /// <summary>
@@ -6,62 +8,67 @@ namespace API.DTOs;
 public class PlayerRatingChartDataPointDTO
 {
     /// <summary>
-    /// Match name
+    /// Match name, if applicable
     /// </summary>
-    public string Name { get; set; } = string.Empty;
+    public string? Name { get; init; }
 
     /// <summary>
-    /// Match id
+    /// Match id, if applicable
     /// </summary>
-    public int? MatchId { get; set; }
+    public int? MatchId { get; init; }
 
     /// <summary>
-    /// osu! match id
+    /// osu! match id, if applicable
     /// </summary>
-    public long? MatchOsuId { get; set; }
+    public long? MatchOsuId { get; init; }
 
     /// <summary>
-    /// Match cost of the player
+    /// Match cost of the player, if applicable
     /// </summary>
-    public double? MatchCost { get; set; }
+    public double? MatchCost { get; init; }
 
     /// <summary>
-    /// Rating of the player before this match occurred
+    /// Rating of the player before the adjustment
     /// </summary>
-    public double RatingBefore { get; set; }
+    public double RatingBefore { get; init; }
 
     /// <summary>
-    /// Rating of the player after this match occurred
+    /// Rating of the player after the adjustment
     /// </summary>
-    public double RatingAfter { get; set; }
+    public double RatingAfter { get; init; }
 
     /// <summary>
     /// Volatility of the player before this match occurred
     /// </summary>
-    public double VolatilityBefore { get; set; }
+    public double VolatilityBefore { get; init; }
 
     /// <summary>
-    /// Volatility of the player after this match occurred
+    /// Volatility of the player after this adjustment
     /// </summary>
-    public double VolatilityAfter { get; set; }
+    public double VolatilityAfter { get; init; }
 
     /// <summary>
-    /// Difference in rating for the player after this match occurred
+    /// Difference in rating between now and the previous adjustment
     /// </summary>
     public double RatingChange => RatingAfter - RatingBefore;
 
     /// <summary>
-    /// Difference in volatility for the player after this match occurred
+    /// Difference in volatility between now and the previous adjustment
     /// </summary>
     public double VolatilityChange => VolatilityAfter - VolatilityBefore;
 
     /// <summary>
-    /// Indicates whether this data point is from a rating change that occurred outside of a match (i.e. decay)
+    /// Ruleset of the adjustment
     /// </summary>
-    public bool IsAdjustment { get; set; }
+    public Ruleset Ruleset { get; init; }
+
+    /// <summary>
+    /// Adjustment type
+    /// </summary>
+    public RatingAdjustmentType RatingAdjustmentType { get; init; }
 
     /// <summary>
     /// Match start time
     /// </summary>
-    public DateTime? Timestamp { get; set; }
+    public DateTime? Timestamp { get; init; }
 }
--- a/API/Repositories/Implementations/ApiMatchRatingStatsRepository.cs
+++ b/API/Repositories/Implementations/ApiMatchRatingStatsRepository.cs
@@ -3,85 +3,92 @@ using API.Repositories.Interfaces;
 using Database;
 using Database.Enums;
 using Database.Repositories.Implementations;
+using Microsoft.EntityFrameworkCore;
 
 namespace API.Repositories.Implementations;
 
-public class ApiMatchRatingStatsRepository(OtrContext context) : MatchRatingStatsRepository(context), IApiMatchRatingStatsRepository
+public class ApiMatchRatingStatsRepository(OtrContext context)
+    : MatchRatingStatsRepository(context), IApiMatchRatingStatsRepository
 {
-    public Task<PlayerRatingChartDTO> GetRatingChartAsync(
+    private readonly OtrContext _context = context;
+
+    public async Task<PlayerRatingChartDTO> GetRatingChartAsync(
         int playerId,
         Ruleset ruleset,
         DateTime? dateMin = null,
         DateTime? dateMax = null
     )
     {
-        // TODO: Rewrite this
-        // // Default date range to min and max values if not provided
-        // dateMin ??= DateTime.MinValue;
-        // dateMax ??= DateTime.MaxValue;
-        //
-        // // Fetch Match Rating Stats and group by Match.StartTime.Date
-        // var matchRatingStats = await _context
-        //     .MatchRatingStats.Where(mrs =>
-        //         mrs.PlayerId == playerId
-        //         && mrs.Match.Tournament.Ruleset == (Ruleset)ruleset
-        //         && mrs.Match.StartTime >= dateMin
-        //         && mrs.Match.StartTime <= dateMax
-        //     )
-        //     .Select(mrs => new
-        //     {
-        //         mrs.Match.StartTime,
-        //         DataPoint = new PlayerRatingChartDataPointDTO
-        //         {
-        //             Name = mrs.Match.Name ?? "<Unknown>",
-        //             MatchId = mrs.MatchId,
-        //             MatchOsuId = mrs.Match.OsuId,
-        //             // MatchCost = mrs.MatchCost,
-        //             RatingBefore = mrs.RatingBefore,
-        //             RatingAfter = mrs.RatingAfter,
-        //             VolatilityBefore = mrs.VolatilityBefore,
-        //             VolatilityAfter = mrs.VolatilityAfter,
-        //             IsAdjustment = false,
-        //             Timestamp = mrs.Match.StartTime
-        //         }
-        //     })
-        //     .ToListAsync();
-        //
-        // // Assuming RatingAdjustments should be grouped by their own timestamp since they may not have a Match.StartTime
-        // var ratingAdjustments = await _context
-        //     .RatingAdjustments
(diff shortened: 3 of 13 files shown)


API.Tests/Services/LeaderboardServiceTests.cs
```diff
@@ -44,7 +46,8 @@ public LeaderboardServiceTests()
             playerRatingsRepository.Object,
             matchStatsRepository.Object,
             playerRepository.Object,
-            tournamentsService
+            tournamentsService,
+            new Mapper(new MapperConfiguration(configuration => configuration.AddProfile(new MapperProfile())))
```
myssto, line 50: Are you meaning to use the `OsuApiClient` mapper profile here?

API/Repositories/Implementations/ApiMatchRatingStatsRepository.cs
```diff
@@ -3,85 +3,68 @@
...
+        var combinedDataPoints = adjustments
+            .Select(mrs => new { mrs.Timestamp.Date, mrs.DataPoint })
+            .GroupBy(x => x.Date)
+            .OrderBy(g => g.Key)
+            .Select(g => g.Select(x => x.DataPoint).ToList())
+            .ToList();
 
         // Prepare and return the DTO
-        return Task.FromResult(new PlayerRatingChartDTO { ChartData = new List<List<PlayerRatingChartDataPointDTO>>() });
+        return new PlayerRatingChartDTO { ChartData = combinedDataPoints };
```
myssto, line 68: Feel like `PlayerRatingChartDTO` is a redundant type since all it is is a 'container' for the datapoints. Could probably be replaced wherever it's used to:
```cs
List<List<PlayerRatingChartDataPointDTO>> RatingChart { get; init; } = [ [ ] ];
```
or something similar

API/Repositories/Implementations/ApiMatchRatingStatsRepository.cs
```diff
@@ -3,85 +3,68 @@
...
+                && ra.Timestamp >= dateMin
+                && ra.Timestamp <= dateMax
+            )
+            .OrderBy(ra => ra.Timestamp)
+            .Select(ra => new
+            {
+                ra.Timestamp,
+                DataPoint = new PlayerRatingChartDataPointDTO
+                {
+                    Name = ra.Match!.Name ?? "<Unknown>",
```
myssto, line 42: Unnecessary nullish coalescence

API/Repositories/Implementations/ApiMatchRatingStatsRepository.cs
```diff
@@ -3,85 +3,68 @@
...
+                    MatchCost = _context.PlayerMatchStats
+                        .Where(pms => pms.PlayerId == ra.PlayerId &&
+                                      pms.MatchId == ra.MatchId)
+                        .Select(pms => pms.MatchCost)
+                        .FirstOrDefault(),
+                    RatingBefore = ra.RatingBefore,
+                    RatingAfter = ra.RatingAfter,
+                    VolatilityBefore = ra.VolatilityBefore,
+                    VolatilityAfter = ra.VolatilityAfter,
+                    IsAdjustment = true,
```
myssto, line 54: Is this a legacy indicator? Why is it always set to `true`?

API/Repositories/Implementations/ApiMatchRatingStatsRepository.cs
```diff
@@ -3,85 +3,68 @@
...
+        dateMin ??= DateTime.MinValue;
+        dateMax ??= DateTime.MaxValue;
+
+        // Fetch Match Rating Stats and group by Match.StartTime.Date
+        var adjustments = await _context
+            .RatingAdjustments
+            .Where(ra =>
+                ra.PlayerId == playerId
+                && ra.AdjustmentType != RatingAdjustmentType.Initial
+                && ra.Match!.Tournament.Ruleset == ruleset
```
myssto, line 32: Smells like nullref? This should use the new `Ruleset` property of the rating adjustment instead right?

API/Repositories/Implementations/ApiMatchRatingStatsRepository.cs
```diff
@@ -3,85 +3,68 @@
...
+        // Default date range to min and max values if not provided
+        dateMin ??= DateTime.MinValue;
+        dateMax ??= DateTime.MaxValue;
+
+        // Fetch Match Rating Stats and group by Match.StartTime.Date
+        var adjustments = await _context
+            .RatingAdjustments
+            .Where(ra =>
+                ra.PlayerId == playerId
+                && ra.AdjustmentType != RatingAdjustmentType.Initial
```
myssto, line 31: Why exclude the initial rating?

API/Services/Implementations/PlayerStatsService.cs
```diff
@@ -274,7 +274,7 @@ DateTime dateMax
             // TODO: Different way of calcing this
             // AverageMatchCostAggregate = ratingStats.Average(x => x.MatchCost),
             HighestRating = adjustments.Max(x => x.RatingAfter),
-            RatingGained = adjustments.Last().RatingAfter - adjustments.First().RatingAfter,
+            RatingGained = adjustments.Last().RatingAfter - adjustments.FirstOrDefault()?.RatingAfter ?? 0,
```
myssto, line 277: Probably bad to use `Last()` over `LastOrDefault()` anywhere (we just fixed that one crash with `First()`). That, and since there is no certainty of sorting here assuming the last adjustment is the most recent is optimistic... This approach probably more consistent:
```cs
RatingGained = adjustments.Sum(ra => ra.RatingDelta),
```
`RatingGained` should probably be renamed to `RatingDelta` or something similar as well

Database/Repositories/Implementations/MatchRatingStatsRepository.cs
```diff
@@ -23,22 +25,18 @@ public async Task<IEnumerable<IEnumerable<RatingAdjustment>>> GetForPlayerAsync(
...
+                && ra.AdjustmentType == RatingAdjustmentType.Match
+                && ra.Ruleset == ruleset
+                && ra.Timestamp >= dateMin
+                && ra.Timestamp <= dateMax
             )
-            .Include(x => x.Match)
-            .ThenInclude(x => x!.Tournament)
-            .GroupBy(x => x.Match!.StartTime.Date)
+            .Include(ra => ra.Match!.Tournament)
+            .GroupBy(ra => ra.Timestamp)
```
myssto, line 36: Is it necessary to group these before returning them? Last time I checked the only place where this is used flattens the list before using the contents anyways /shrug

myssto: Probably makes more sense to let the caller of this function group them how they want to either way instead of this

Database/Repositories/Implementations/PlayerRatingsRepository.cs
```diff
@@ -24,33 +24,31 @@ public async Task<IEnumerable<PlayerRating>> GetAsync(long osuPlayerId)
...
     {
-        var ls = new List<PlayerRating>();
-        foreach (PlayerRating stat in playerRatings)
+        var results = await _context.PlayerRatings
+            .Include(pr => pr.Adjustments)
+            .Where(pr => pr.PlayerId == playerId && pr.Ruleset == ruleset)
+            .Select(pr => new
+            {
+                PlayerRating =
+                    pr, // TODO: Should probably include all RatingAdjustments regardless of type and have web not display initial
```
myssto, line 35: Similar to the above comment, I don't see why we are ever excluding the initial rating to begin with... Besides that shouldn't be API's responsibility. Whoever is requesting the data can do what they want with it and we should try to include the full dataset where reasonable

Database/Repositories/Implementations/PlayerRatingsRepository.cs
```diff
@@ -24,33 +24,31 @@ public async Task<IEnumerable<PlayerRating>> GetAsync(long osuPlayerId)
...
+                Adjustments = pr.Adjustments.Where(ra => ra.AdjustmentType != RatingAdjustmentType.Initial)
+                    .OrderBy(ra => ra.Timestamp)
+                    .ToList()
+            })
```
myssto, line 39: What is with this projection? You can do a filtered include on the adjustments if that is the goal

myssto (changes requested):

API/Repositories/Implementations/ApiMatchRatingStatsRepository.cs
hburn7: I avoided that originally because to me it feels harder to maintain, and there's a non-zero chance we expand this DTO in the future.

API/Repositories/Implementations/ApiMatchRatingStatsRepository.cs
hburn7: Legacy yeah, this is from before we had `RatingAdjustmentType` I'm pretty sure.

API.Tests/Services/LeaderboardServiceTests.cs
hburn7: I don't think so, the mapper here is used [here](https://github.com/osu-tournament-rating/otr-api/blob/9275faf51a161372779589ac507e7f99495bcb84/API/Services/Implementations/PlayerRatingsService.cs#L78):

```cs
Adjustments = mapper.Map<ICollection<RatingAdjustmentDTO>>(currentStats.Adjustments)
```

API/Repositories/Implementations/ApiMatchRatingStatsRepository.cs
hburn7: Yes thank you

hburn7: Also going to remove this line: `&& ra.AdjustmentType != RatingAdjustmentType.Initial` as I think that's something the frontend should handle.

API/Services/Implementations/PlayerStatsService.cs
hburn7: Renaming would make this a breaking change so would rather not do that yet. But yes sum of rating delta sounds like it will achieve what we want.

Database/Repositories/Implementations/PlayerRatingsRepository.cs
hburn7: The projection here is to execute a subquery which fetches the `MatchCost`. You can do subqueries by projecting via `Select` but not by `Where`.

Database/Repositories/Implementations/MatchRatingStatsRepository.cs
hburn7: Agreed the repository shouldn't do the grouping here.

Database/Repositories/Implementations/PlayerRatingsRepository.cs
hburn7: yeah this was kinda scuffed

API/DTOs/PlayerRatingChartDataPointDTO.cs
```diff
@@ -6,62 +8,67 @@ namespace API.DTOs;
 public class PlayerRatingChartDataPointDTO
 {
     /// <summary>
-    /// Match name
+    /// Match name, if applicable
```
myssto, line 11: Probably out of scope for this PR but this application right here is exactly where I would love to use a `MatchCompactDTO` instead of spreading a ton of nullable related properties.

API/Repositories/Implementations/ApiMatchRatingStatsRepository.cs
```diff
@@ -3,85 +3,65 @@
...
+        List<IGrouping<DateTime?, PlayerRatingChartDataPointDTO>> chartPoints = await _context
+            .RatingAdjustments
+            .Include(ra => ra.Match)
+            .Where(ra =>
+                ra.PlayerId == playerId
+                && ra.Ruleset == ruleset
+                && ra.Timestamp >= dateMin
+                && ra.Timestamp <= dateMax
+            )
+            .OrderBy(ra => ra.Timestamp)
```
myssto, line 36: This `OrderBy` is unnecessary if you do it again before executing the query at the end

API/Services/Implementations/PlayerStatsService.cs
```diff
@@ -260,35 +260,43 @@ DateTime dateMax
...
         }
 
+        var offset = adjustments.FirstOrDefault(ra => ra.AdjustmentType == RatingAdjustmentType.Initial)?.RatingAfter;
+
+        if (offset is null)
+        {
+            logger.LogWarning("Initial rating adjustment not found for player, stats will not be completely accurate " +
+                            " [Id: {PlayerId}]", id);
+        }
+
```
myssto, line 279: I think I understand what this is doing, but can you explain why its necessary?

Database/Repositories/Implementations/PlayerRatingsRepository.cs
```diff
@@ -24,33 +24,17 @@ public async Task<IEnumerable<PlayerRating>> GetAsync(long osuPlayerId)
...
-        }
+        PlayerRating? result = await _context.PlayerRatings
+            .Include(pr => pr.Adjustments)
+            .Where(pr => pr.PlayerId == playerId && pr.Ruleset == ruleset)
+            .FirstOrDefaultAsync();
 
-        await _context.PlayerRatings.AddRangeAsync(ls);
-        return await _context.SaveChangesAsync();
+        return result;
     }
```
myssto, line 35: Convert to arrow function that returns the result of the query

myssto (changes requested):
Looking great just small stuff

API/DTOs/PlayerRatingChartDataPointDTO.cs
hburn7: Can make an issue and implement after merging maybe?

API/Services/Implementations/PlayerStatsService.cs
hburn7: Basically if you sum all of the `RatingAfter` values without subtracting from the initial rating you end up with a large negative value.

Will modify this implementation slightly because offset isn't actually required. If a user selects a timespan of say 30 days there won't be an initial rating. But, if the initial rating is included in the result set, we need to account for it.

myssto (approved):

PR #571, opened 2024-12-24
Upgrade various packages to latest version

hburn7:

PR #572, opened 2024-12-24
Fix duplicate tournament stat entries

hburn7:
When re-running the processor/dataworker flow multiple times, tournament stats will be generated and attempted to be inserted multiple times. This PR adds a check to skip adding new tournament stats if they already exist for the player, for the current tournament.

Depends on:

- [x] #571 

@myssto can you confirm if this is a safe implementation? I think so because tournament stats don't seem to be variable between runs.


myssto:
> I think so because tournament stats don't seem to be variable between runs.

They are variable between runs because they take a total rating delta over all matches played. They need to be truncated before each ratings processor run

hburn7:
Fixed in https://github.com/osu-tournament-rating/otr-processor/pull/86

Issue #575, opened 2025-01-05
Update filtering parameters on existing endpoints


hburn7:
API Features:
- [x] `GET` `/matches/{id}` should have the same `verified` query param functionality as tournaments https://github.com/osu-tournament-rating/otr-api/pull/576
- [x] Add `GET` `/games/{id}` with the same `verified` query param functionality https://github.com/osu-tournament-rating/otr-api/pull/577
- [x] Add `GET` `/scores/{id}` (doesnt need query param since no child filtering) https://github.com/osu-tournament-rating/otr-api/pull/578

PR #576, opened 2025-01-06
Add support for `verified` filter on `/matches/{id}`, include `Players` on `MatchDTO`, and other fixes

hburn7:
Part of #575 

- Fixed bug with `DELETE /matches/{id}` - `MatchWinRecord` had a delete behavior of `SetNull`. However, the foreign key `FK_match_win_records_matches_match_id` is required to be non-null.
- Added `bool verified` to `GET /matches/{id}`
- Added supporting changes to allow inclusion of verified children if specified
- Include `PlayerCompactDTO` of `Players` on `MatchDTO`

### DIFF
--- a/API/Configurations/MapperProfile.cs
+++ b/API/Configurations/MapperProfile.cs
@@ -21,7 +21,9 @@ public class MapperProfile : Profile
         CreateMap<GameScore, GameScoreDTO>().ForMember(x => x.Misses, opt => opt.MapFrom(y => y.CountMiss));
 
         CreateMap<Match, MatchDTO>()
-            .ForMember(x => x.Ruleset, opt => opt.MapFrom(x => x.Tournament.Ruleset));
+            .ForMember(x => x.Ruleset, opt => opt.MapFrom(x => x.Tournament.Ruleset))
+            .ForMember(x => x.Players, opt => opt.Ignore());
+
         CreateMap<Match, MatchSubmissionStatusDTO>();
         CreateMap<Match, MatchCreatedResultDTO>()
             .MapAsCreatedResult()
--- a/API/Configurations/MapperProfile.cs
+++ b/API/Configurations/MapperProfile.cs
@@ -38,7 +38,8 @@ public class MapperProfile : Profile
 
         CreateMap<RatingAdjustment, RatingAdjustmentDTO>();
 
-        CreateMap<Player, PlayerCompactDTO>();
+        CreateMap<Player, PlayerCompactDTO>()
+            .ForMember(x => x.UserId, opt => opt.MapFrom(y => y.User!.Id));
         CreateMap<PlayerOsuRulesetData, PlayerOsuRulesetDataDTO>();
 
         CreateMap<Tournament, TournamentCompactDTO>();
--- a/API/Controllers/MatchesController.cs
+++ b/API/Controllers/MatchesController.cs
@@ -30,6 +30,7 @@ public partial class MatchesController(IMatchesService matchesService, IAdminNot
     /// Get a match
     /// </summary>
     /// <param name="id">Match id</param>
+    /// <param name="verified">Whether the match and all child navigations must be verified</param>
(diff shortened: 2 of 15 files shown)


API/Configurations/MapperProfile.cs
```diff
@@ -36,7 +38,8 @@ public MapperProfile()
 
         CreateMap<RatingAdjustment, RatingAdjustmentDTO>();
 
-        CreateMap<Player, PlayerCompactDTO>();
+        CreateMap<Player, PlayerCompactDTO>()
+            .ForMember(x => x.UserId, opt => opt.MapFrom(y => y.User!.Id));
```
myssto, line 42: `Player.User` is an optional navigation right? This may have been my mistake but `PlayerCompactDTO.UserId` should be nullable if it isn't. `opt.MapFrom(y => y.User!.Id)` smells like null-ref

API/Services/Implementations/MatchesService.cs
```diff
@@ -74,4 +95,16 @@ public async Task<bool> ExistsAsync(int id) =>
 
     public async Task DeleteAsync(int id) =>
         await matchesRepository.DeleteAsync(id);
+
+    private async Task<ICollection<PlayerCompactDTO>> GetPlayerCompactsAsync(MatchDTO match)
+    {
+        IEnumerable<int> playerIds = match.Games
+            .Select(x => x.Scores.Select(y => y.PlayerId))
+            .SelectMany(x => x)
```
myssto, line 103:
```suggestion
            .SelectMany(g => g.Scores.Select(s => s.PlayerId))
```
Think this can be condensed, also lambda var names

Database/Repositories/Implementations/PlayersRepository.cs
```diff
@@ -22,6 +22,11 @@ public async Task<IEnumerable<Player>> GetByOsuIdAsync(IEnumerable<long> osuIds)
         return (await _context.Players.Where(p => remainingIds.Contains(p.OsuId)).ToListAsync()).Concat(result);
     }
 
+    public new async Task<ICollection<Player>> GetAsync(IEnumerable<int> ids) =>
```
myssto, line 25: `public new async` ?
Did you mean `override` or is this some syntax I don't know about?

API/Controllers/MatchesController.cs
```diff
@@ -36,9 +36,9 @@ public async Task<IActionResult> ListAsync([FromQuery] MatchRequestQueryDTO requ
     [Authorize(Roles = $"{OtrClaims.Roles.User}, {OtrClaims.Roles.Client}")]
     [ProducesResponseType(StatusCodes.Status404NotFound)]
     [ProducesResponseType<MatchDTO>(StatusCodes.Status200OK)]
-    public async Task<IActionResult> GetAsync(int id)
+    public async Task<IActionResult> GetAsync(int id, [FromQuery] bool verified = true)
```
myssto, line 39: New parameter needs to be documented

Database/Repositories/Implementations/RepositoryBase.cs
```diff
@@ -43,6 +43,12 @@ public virtual async Task<IEnumerable<T>> CreateAsync(IEnumerable<T> entities)
 
     public virtual async Task<T?> GetAsync(int id) => await _context.Set<T>().FindAsync(id);
 
+    public async Task<ICollection<T>> GetAsync(IEnumerable<int> ids) =>
```
myssto, line 46:
```suggestion
    public virtual async Task<ICollection<T>> GetAsync(IEnumerable<int> ids) =>
```
Ah ok the previous comment now makes sense.

myssto (changes requested):
big stuff

API/Configurations/MapperProfile.cs
hburn7: I did test and null-ref doesn't happen here, it behaves like an include. The UserId here does get populated if a user is present and is null otherwise.

myssto (approved):

PR #577, opened 2025-01-06
Add `GET /games/{id}` with `verified` parameter

hburn7:
Depends on: #576 
Part of: #575 

- Adds `GET /games/{id}` endpoint with `verified` parameter.

PR #578, opened 2025-01-06
Add `GET /gamescores/{id}` with `verified` parameter

hburn7:
Depends on #577 
Part of #575 

- Adds `GET /gamescores/{id}` with `verified` parameter.


myssto:
> Add GET /scores/{id} (doesnt need query param since no child filtering)

Everyone point and laugh! But on a real note, query param doesn't need to be here because a score has no children to be filtered

hburn7:
right, derp lol

myssto (changes requested):
Should be updated to behave like the previous pulls

myssto (approved):

Issue #580, opened 2025-01-10
Include more data on existing tournament item DTOs


myssto:
Tournament items returned from the API should include all of the stats that we generate. Some of the following DTOs are already implemented but as part of this process they should be double checked to ensure code quality and proper documentation.

`TournamentDTO`
- [x] `PlayerTournamentStatsDTO[] PlayerStats`
- [x] `PlayerCompactDTO[] Players` (loaded in the same manner as on the MatchDTO)

`MatchDTO`
- [x] `RatingAdjustmentDTO[] RatingAdjustments`
- [x] `PlayerMatchStatsDTO[] PlayerStats`
- [x] `MatchWinRecordDTO? WinRecord`

`GameDTO`
- [x] `GameWinRecordDTO? WinRecord`

`GameScoreDTO`
- [x] Include `int? Placement` on the existing `GameScoreDTO`

hburn7:
`PlayerTournamentStatsDTO` is currently designed to be attached to a player's statistics object. Therefore it is designed to be from the perspective of the player. I think having a new DTO, such as `TournamentPlayerStatsDTO` would be appropriate. Then we can design it to be from the perspective of the tournament which will work better imo.

`RatingAdjustmentDTO` currently does not feature a `PlayerId` field. I think doing an adhoc `Dictionary`, such as `Dictionary<int, RatingAdjustmentDTO>` would be fine here.

yay/nay/wanna fight?

myssto:
> yay/nay/wanna fight?

I choose violence

>PlayerTournamentStatsDTO is currently designed to be attached to a player's statistics object. Therefore it is designed to be from the perspective of the player. I think having a new DTO, such as TournamentPlayerStatsDTO would be appropriate. Then we can design it to be from the perspective of the tournament which will work better imo.

The `PlayerTournamentStats` entity is 1000% not *only* designed to be viewed from the perspective of the player. It is simply an aggregate of a player performance over an entire tournament. Currently we use `PlayerTournamentMatchCostDTO` (which probably just needs to be removed in favor of a new `PlayerTournamentStatsDTO` based on the entity directly) to give a player performance summary for the dashboard but it can be used anywhere. I am imagining an individual tournament page can have a list of participating players, displayed in order of average match cost, average score, etc.

>RatingAdjustmentDTO currently does not feature a PlayerId field. I think doing an adhoc Dictionary, such as Dictionary<int, RatingAdjustmentDTO> would be fine here.

[...]

PR #581, opened 2025-01-13
Additions to `GameDTO` and `GameScoreDTO`

### LINKED ISSUE #579: Add Players field to GameDTO
Copy the implementation from MatchDTO

myssto:
Closes #579
Part of #580

Adds additional information to `GameDTO` and `GameScoreDTO`
- `GameScore` statistics
- `GameWinRecord` updated and included on `GameDTO`
- Load `PlayerCompactDTO`s when requesting a game

### DIFF
--- a/API/DTOs/GameDTO.cs
+++ b/API/DTOs/GameDTO.cs
@@ -80,6 +80,19 @@ public class GameDTO
     public required BeatmapDTO Beatmap { get; init; }
 
     /// <summary>
+    /// Win record
+    /// </summary>
+    public GameWinRecordDTO? WinRecord { get; init; }
+
+    /// <summary>
+    /// All participating players
+    /// </summary>
+    /// <remarks>
+    /// Will only be populated if the game is the highest order of entity requested
+    /// </remarks>
+    public ICollection<PlayerCompactDTO> Players { get; set; } = [];
+
+    /// <summary>
     /// All associated admin notes
     /// </summary>
     [SuppressMessage("ReSharper", "CollectionNeverUpdated.Global")]
--- a/API/DTOs/GameDTO.cs
+++ b/API/DTOs/GameDTO.cs
@@ -7,82 +7,87 @@ namespace API.DTOs;
 /// <summary>
 /// Represents a single game (osu! beatmap) played in a match
 /// </summary>
-[SuppressMessage("ReSharper", "PropertyCanBeMadeInitOnly.Global")]
 public class GameDTO
 {
     /// <summary>
     /// Primary key
     /// </summary>
-    public int Id { get; set; }
+    public int Id { get; init; }
 
     /// <summary>
     /// The ruleset
     /// </summary>
-    public Ruleset Ruleset { get; set; }
+    public Ruleset Ruleset { get; init; }
 
     /// <summary>
     /// The scoring type used
     /// </summary>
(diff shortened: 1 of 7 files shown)


API/DTOs/GameDTO.cs
```diff
@@ -1,88 +1,108 @@
...
+    public TeamType TeamType { get; init; }
 
     /// <summary>
     /// The mods enabled
     /// </summary>
-    public Mods Mods { get; set; }
+    public Mods Mods { get; init; }
+
+    /// <summary>
+    /// Denotes if the mod setting was "free mod"
```
hburn7, line 39:
```suggestion
    /// Denotes if the mod setting is "free mod"
```

API/DTOs/GameDTO.cs
```diff
@@ -1,88 +1,108 @@
...
     /// Timestamp of the end of the game
     /// </summary>
-    public DateTime? EndTime { get; set; }
+    public DateTime? EndTime { get; init; }
 
     /// <summary>
     /// The beatmap played
     /// </summary>
-    public BeatmapDTO? Beatmap { get; set; }
+    public required BeatmapDTO Beatmap { get; init; }
```
hburn7, line 81: Is `required` something you added or is this a formatting thing? Haven't we been using `[Required]` instead?

API/Services/Implementations/GamesService.cs
```diff
@@ -38,4 +48,10 @@ public async Task<bool> ExistsAsync(int id) =>
 
     public async Task DeleteAsync(int id) =>
         await gamesRepository.DeleteAsync(id);
+
+    private async Task<ICollection<PlayerCompactDTO>> GetPlayerCompactsAsync(GameDTO game)
+    {
+        IEnumerable<int> playerIds = game.Scores.Select(s => s.PlayerId);
```
hburn7, line 54: Does this need a `.Distinct` call?

myssto: Yes, thanks good catch

API/DTOs/GameDTO.cs
myssto: Just took a look and it's being marked non-null correctly. iirc the attribute is specific for input (like query params / model binding).

hburn7: Can you try double checking without `required`? The type has changed from nullable to non-nullable so I'm wondering if `required` is doing what you expect it to. Don't mean to drag this out over something so small, just want to make sure we understand what `required` is doing here.

According to MSDN, `required` is only used when a field must be populated by an object initializer (see image). For DTOs, we are using AutoMapper to take care of initialization in most cases (and I believe this case is covered by AutoMapper as well).

image

myssto: No worries about dragging things out we're learning either way :P

I understand what you're getting at, but the whole point is to denote that the property is non-null. In fairness using `= null!` will also accomplish this and is more consistent with what we use everywhere else. If `required` is removed, it rightfully warns for the inspection `Non-nullable property 'Beatmap' is uninitialized. Consider declaring the property as nullable.` where the possible fixes are to add a `required` keyword or use null assertion.

myssto: I'll change it to use null assertion.

hburn7 (approved):

PR #590, opened 2025-02-02
include new beatmap navigations in queries

myssto:


Database/Repositories/Implementations/TournamentsRepository.cs
```diff
@@ -301,6 +301,12 @@ private IQueryable<Tournament> TournamentsBaseQuery()
             .Include(e => e.Matches)
             .ThenInclude(m => m.Games)
             .ThenInclude(g => g.Beatmap)
+            .ThenInclude(b => b != null ? b.BeatmapSet : null)
+            .ThenInclude(bs => bs != null ? bs.Creator : null)
+            .Include(e => e.Matches)
+            .ThenInclude(m => m.Games)
+            .ThenInclude(g => g.Beatmap)
+            .ThenInclude(b => b != null ? b.Creators : null)
```
hburn7, line 309: No need for null checks here:

> When dealing with optional relationships, it's possible to encounter compiler warnings where an actual null reference exception would be impossible. When translating and executing your LINQ queries, EF Core guarantees that if an optional related entity does not exist, any navigation to it will simply be ignored, rather than throwing. However, the compiler is unaware of this EF Core guarantee, and produces warnings as if the LINQ query were executed in memory, with LINQ to Objects. As a result, it is necessary to use the null-forgiving operator (!) to inform the compiler that an actual null value isn't possible[.]

https://learn.microsoft.com/en-us/ef/core/miscellaneous/nullable-reference-types#navigating-and-including-nullable-relationships

Database/Utilities/Extensions/QueryExtensions.cs
```diff
@@ -270,24 +270,27 @@ public static IQueryable<Match> IncludeChildren(this IQueryable<Match> query, bo
...
+            .ThenInclude(gs => gs.AdminNotes)
+            .Include(m => m.Games)
             .ThenInclude(g => g.AdminNotes)
             .Include(m => m.Games)
             .ThenInclude(g => g.Beatmap)
+            .ThenInclude(b => b != null ? b.BeatmapSet : null)
+            .ThenInclude(bs => bs != null ? bs.Creator : null)
+            .Include(m => m.Games)
+            .ThenInclude(g => g.Beatmap)
+            .ThenInclude(b => b != null ? b.Creators : null)
```
hburn7, line 289: See above

hburn7 (changes requested):

Database/Repositories/Implementations/TournamentsRepository.cs
myssto: oh cool thanks for this

myssto:
Tournaments repo could use some query extensions... lol

hburn7:
Yeahhhh.. it's alright for now I guess. We can make a tracking issue.

hburn7 (approved):

PR #600, opened 2025-02-13
rewrite the data parser to support new beatmap structure

myssto:


hburn7 (approved):

hburn7:
Re: formatting CI..

I personally am leaning towards disabling this rule. I'm old school and like the `.ToList` syntax. What about you?

PR #610, opened 2025-02-22
Exclude empty matches from tournaments threshold calculation in checks

### LINKED ISSUE #592: Do not include `NoGames` matches in tournament `NotEnoughVerifiedMatches` threshold
We have a rule which pre-rejects tournaments which have 20%+ rejected matches. If a match has the `NoGames` rejection reason (and no other rejection reason), it should be excluded from this calculation.

- [ ] In `MatchGameCountCheck`, update the logic to exclude matches with `NoGames` from the count of valid games.

majorro:
Closes #592


majorro:
This needs some fix for randomized tests - making it static or adding hacky clauses

hburn7 (commented):
Logic change looks good, I would appreciate an additional test for this specific behavior though.

hburn7 (approved):

PR #620, opened 2025-03-10
Nullable start and end times for matches and tournaments

### LINKED ISSUE #614: Tournament start time set to `2007-09-16` inadvertently
There is a bug where tournaments have a faulty `StartTime` value. This value is controlled by the `TournamentDataProcessor`. All tournaments should have a valid `StartTime` which is derived from the `Matches` list.

There is probably some logic where matches are having their `StartTime` value set to this default if there are no games or if there are other issues with the match. Instead of looking directly at the first match's start time, we need to have a smarter check which looks for the first reasonable start time and use that instead.

# Tasks

- [ ] Change the type of `Match.StartTime` to be nullable (`DateTime?`).
- [ ] Modify logic such that a `Match` with valid `Game`s should never have a `null` `StartTime` value.
- [ ] Modify logic such that a `Tournament` will never have an invalid start time if it has any matches with a non-null start time. The earliest known match start time should be used.

majorro:
Closes #614 

- `StartTime` and `EndTime` in `Tournament` and `Match` are now nullable instead of using placeholder
- Consistent border dates determining logic in data processors
- Data migration putting dates using the same logic, null otherwise


majorro:
Failing checks opentelemetry thing is fixed in !610 btw

DataWorkerService.Tests/AutomationChecks/Matches/MatchEndTimeCheckTests.cs
```diff
@@ -27,11 +27,11 @@ public void Check_PassesWhenExpected(DateTime endTime, bool expectedPass)
     }
 
     [Fact]
-    public void Check_GivenDefaultDateTime_Fails()
+    public void Check_GivenEmptyDateTime_Fails()
```
hburn7, line 30:
```suggestion
    public void Match_IsRejected_WhenNoEndTime()
```

DataWorkerService/Services/Implementations/OsuApiDataParserService.cs
```diff
@@ -81,7 +80,7 @@ public async Task ParseMatchAsync(Match match, MultiplayerMatch apiMatch)
                 {
                     // Scale up EZ scores
                     Score = mpScore.Mods.HasFlag(Mods.Easy)
```
hburn7, line 82: @myssto Do we need to add an audit entry here for this score override?

DataWorkerService.Tests/SharedTestData.cs
```diff
@@ -58,8 +58,8 @@ public class EndTimeTestData : IEnumerable<object[]>
     {
         public IEnumerator<object[]> GetEnumerator()
         {
-            yield return [SeededDate.Placeholder, false];
-            yield return [DateTime.MinValue, false];
+            yield return [SeededDate.Placeholder, true];
```
hburn7, line 61: Do we need something like `yield return [null, false]` here? Everything is `true` now. @myssto ?

DataWorkerService/Services/Implementations/OsuApiDataParserService.cs
```diff
@@ -81,7 +80,7 @@ public async Task ParseMatchAsync(Match match, MultiplayerMatch apiMatch)
...
-                        ? (int)(mpScore.Score * 1.75)
+                        ? (int)(mpScore.Score * 1.75) // TODO: parametrize
```
hburn7, line 83:
```suggestion
                        ? (int)(mpScore.Score * 1.75)
```

Can convert to a constant instead at the top of the file.

hburn7 (changes requested):
Normally I would not allow a one-way migration like this (where there is no inverse of the SQL in the Down() method of the migration), but it's actually not needed in this case. I've tested and verified the behavior of the migrations and the timestamp update sql. Nice work on this.

I also need an additional unit test which covers the new `TournamentDataProcessor` functionality.

DataWorkerService.Tests/SharedTestData.cs
majorro: theres an additional test for this case, if add null case here, then seeding workaround would be needed since it generates dates for nulls

majorro:
> but it's actually not needed in this case

yeah, they just will be filled with default values which are placeholders

DataWorkerService.Tests/AutomationChecks/Matches/MatchEndTimeCheckTests.cs
majorro: why? all other tests are named in that way. but ye rejection reason should be added to name

DataWorkerService/Services/Implementations/OsuApiDataParserService.cs
majorro: sure

DataWorkerService.Tests/AutomationChecks/Matches/MatchEndTimeCheckTests.cs
hburn7: This name is more descriptive of what the test is actually doing. There are probably a lot of poorly written test names.

hburn7 (approved):
Looks okay I think.

hburn7:
nvm, let's get this in.

PR #621, opened 2025-03-10
Games integrity checks improvements

### LINKED ISSUE #615: LobbySizeMIsmatch only checks for total number of scores, not per team number
In the data worker, LobbySizeMismatch is defined as a rejection reason if the total number of scores divided by 2 is not equal to the team size. However this can fail in cases with uneven team sizes that still add up to team size x 2. 

e. g. https://otr.stagec.xyz/matches/255 
See warmup in this match.

majorro:
Closes #615 

- `LobbySizeMismatch` now considers teams inequality cases
- New warning flag `RostersNotUnique` and check `MatchTeamsIntegrityCheck` for player appearing in different teams among match games
- Added some tests
- Extracted `RostersHelper`


DataWorkerService/AutomationChecks/Matches/MatchTeamsIntegrityCheck.cs
```diff
@@ -0,0 +1,38 @@
...
+            .Where(x => x.VerificationStatus.IsPreVerifiedOrVerified())];
+
+        var teamRedPlayers = validScores
+            .Where(x => x.Team == Team.Red)
+            .Select(x => x.PlayerId)
+            .ToHashSet();
+
+        IEnumerable<int> teamBluePlayers = validScores
+            .Where(x => x.Team == Team.Blue)
+            .Select(x => x.PlayerId);
```
hburn7, line 29: Idk how I feel about grouping by red/blue specifically here. I would look into the `Roster` logic and apply that here instead.

DataWorkerService/AutomationChecks/Games/GameScoreCountCheck.cs
```diff
@@ -27,8 +29,19 @@ protected override bool OnChecking(Game entity)
             return false;
         }
 
-        // Number of scores matches expected team size
-        if (validScoresCount % 2 == 0 && validScoresCount / 2 == entity.Match.Tournament.LobbySize)
+        if (entity.TeamType is TeamType.TeamVs or TeamType.TagTeamVs)
```
hburn7, line 32: I don't think we support TagTeamVs. @cytusine0 confirm please

DataWorkerService/AutomationChecks/Games/GameScoreCountCheck.cs
```diff
@@ -27,8 +29,19 @@ protected override bool OnChecking(Game entity)
...
+        {
+            var redValidScoresCount = validScores.Count(gs => gs.Team == Team.Red);
+            var blueValidScoresCount = validScores.Count(gs => gs.Team == Team.Blue);
+
+            if (redValidScoresCount == blueValidScoresCount &&
+                redValidScoresCount == entity.Match.Tournament.LobbyTeamSize &&
+                redValidScoresCount + blueValidScoresCount == validScoresCount)
+            {
+                return true;
+            }
```
hburn7, line 42: Same thing, look into roster logic instead of using the colors. You could probably extract some common Roster utilities so that we're not reusing logic in multiple places.

hburn7 (changes requested):
There are still some instances of `is VerificationStatus.PreVerified or VerificationStatus.Verified` being used in the codebase. I'd make sure to update all of those too.

DataWorkerService/AutomationChecks/Games/GameScoreCountCheck.cs
cytusine0: yup can confirm 👍  

are HeadToHead team types already fully converted by this point in the logic? if not, maybe would be good to also have a case for that, checking that the total number of players is twice `LobbySize`

majorro:
> There are still some instances of `is VerificationStatus.PreVerified or VerificationStatus.Verified` being used in the codebase. I'd make sure to update all of those too.

I was unable to find any... Maybe you've seen them in adjacent prs?

DataWorkerService/AutomationChecks/Matches/MatchTeamsIntegrityCheck.cs
majorro: yeah, roster generation just groups by team, but even if there would be a case where players exist in a game with all 3 enum values, it vould fail in `GameScoreCountCheck`

DataWorkerService/AutomationChecks/Games/GameScoreCountCheck.cs
majorro: > I don't think we support TagTeamVs. @cytusine0 confirm please

actually i added it "just in case"

majorro: > checking that the total number of players is twice `LobbySize`

it checked in https://github.com/osu-tournament-rating/otr-api/blob/cd8ca2bd8f85140d3173f7a1f8520c64214b695d/DataWorkerService/AutomationChecks/Games/GameScoreCountCheck.cs#L39

DataWorkerService/AutomationChecks/Games/GameScoreCountCheck.cs
majorro: done

DataWorkerService/AutomationChecks/Games/GameScoreCountCheck.cs
majorro: removed these checks

DataWorkerService/AutomationChecks/Matches/MatchTeamsIntegrityCheck.cs
majorro: done

hburn7:
I want to brainstorm some alternative names for this. We could use `Format`, `FormatSize`, `ActiveLobbySize`, etc. @cytusine0 @myssto

hburn7:
> > There are still some instances of `is VerificationStatus.PreVerified or VerificationStatus.Verified` being used in the codebase. I'd make sure to update all of those too.
> 
> I was unable to find any... Maybe you've seen them in adjacent prs?

Huh yeah I might've been looking at another branch, nevermind

hburn7:
@majorro Please rename `LobbyTeamSize` to `TeamLobbySize`. That should be good. I do agree that `LobbySize` is ambiguous and not actually correct.

majorro:
> @majorro Please rename `LobbyTeamSize` to `TeamLobbySize`. That should be good. I do agree that `LobbySize` is ambiguous and not actually correct.

done 👌

DataWorkerService/AutomationChecks/Matches/MatchTeamsIntegrityCheck.cs
```diff
@@ -0,0 +1,47 @@
...
+            return true;
+        }
+
+        for (var i = 0; i < playerIdsPerRoster.Length; ++i)
+        {
+            for (var j = i + 1; j < playerIdsPerRoster.Length; ++j)
+            {
+                if (playerIdsPerRoster[i].Overlaps(playerIdsPerRoster[j]))
+                {
+                    entity.WarningFlags |= MatchWarningFlags.SamePlayerAmongTeams;
```
hburn7, line 39: Let's rename this to `RostersNotUnique`

DataWorkerService/AutomationChecks/Games/GameScoreCountCheck.cs
```diff
@@ -27,8 +30,19 @@ protected override bool OnChecking(Game entity)
             return false;
         }
 
-        // Number of scores matches expected team size
-        if (validScoresCount % 2 == 0 && validScoresCount / 2 == entity.Match.Tournament.LobbySize)
+        if (entity.TeamType is TeamType.TeamVs)
```
hburn7, line 33: This is handled in a separate check, no need to do that here.

DataWorkerService/AutomationChecks/Games/GameScoreCountCheck.cs
```diff
@@ -27,8 +30,19 @@ protected override bool OnChecking(Game entity)
...
+            var playerCountPerTeam = rosters.Select(x => x.Roster.Length).ToArray();
+
+            if (playerCountPerTeam.Length > 1 && // more than one team
+                playerCountPerTeam.All(x => x == playerCountPerTeam[0]) && // all counts are equal
+                playerCountPerTeam[0] == entity.Match.Tournament.LobbySize) // all counts are correct
+            {
+                return true;
+            }
+        }
+        else if (validScoresCount % 2 == 0 && validScoresCount / 2 == entity.Match.Tournament.LobbySize)
```
hburn7, line 45: Put this on line 33

DataWorkerService/AutomationChecks/Games/GameScoreCountCheck.cs
```diff
@@ -27,8 +30,19 @@ protected override bool OnChecking(Game entity)
...
+        {
+            ICollection<GameRoster> rosters = RostersHelper.GenerateRosters(validScores);
+            var playerCountPerTeam = rosters.Select(x => x.Roster.Length).ToArray();
+
+            if (playerCountPerTeam.Length > 1 && // more than one team
+                playerCountPerTeam.All(x => x == playerCountPerTeam[0]) && // all counts are equal
+                playerCountPerTeam[0] == entity.Match.Tournament.LobbySize) // all counts are correct
+            {
+                return true;
+            }
```
hburn7, line 43: Sorry for the lack of constructive feedback but this code looks kind of bad. Can you find another way to do this? The zero-indexing and all just feels wrong for this.

hburn7 (changes requested):

DataWorkerService/AutomationChecks/Games/GameScoreCountCheck.cs
SourMongoose: Should be able to simplify this using `.all()` or similar

DataWorkerService/AutomationChecks/Matches/MatchTeamsIntegrityCheck.cs
majorro: sure

DataWorkerService/AutomationChecks/Games/GameScoreCountCheck.cs
majorro: sure

DataWorkerService/AutomationChecks/Games/GameScoreCountCheck.cs
majorro: yeah, actually i think some checks decoupling is a good idea in a long run, but okay

DataWorkerService/AutomationChecks/Games/GameScoreCountCheck.cs
majorro: thats why i put comments here xd, okay, rewrote, it went cleaner with adding score count checks before 👍

DataWorkerService/AutomationChecks/Games/GameScoreCountCheck.cs
```diff
@@ -30,22 +30,15 @@ protected override bool OnChecking(Game entity)
...
-        if (entity.TeamType is TeamType.TeamVs)
+        if (validScoresCount % 2 == 0 && validScoresCount / 2 == entity.Match.Tournament.LobbySize)
         {
             ICollection<GameRoster> rosters = RostersHelper.GenerateRosters(validScores);
-            var playerCountPerTeam = rosters.Select(x => x.Roster.Length).ToArray();
 
-            if (playerCountPerTeam.Length > 1 && // more than one team
-                playerCountPerTeam.All(x => x == playerCountPerTeam[0]) && // all counts are equal
-                playerCountPerTeam[0] == entity.Match.Tournament.LobbySize) // all counts are correct
+            if (rosters.DistinctBy(x => x.Roster.Length).Count() == 1)
```
SourMongoose, line 37: Will this work if there is one team of 4 players for a 2v2 lobby?

Common/Enums/Enums/Verification/MatchWarningFlags.cs
```diff
@@ -28,5 +28,11 @@ public enum MatchWarningFlags
...
     /// of <see cref="GameRejectionReason.BeatmapNotPooled"/> outside of the first two <see cref="Game"/>s
     /// </summary>
-    UnexpectedBeatmapsFound = 1 << 2
+    UnexpectedBeatmapsFound = 1 << 2,
+
+    /// <summary>
+    /// The <see cref="Database.Entities.Match"/> has 1 or more <see cref="Database.Entities.Player"/>s that played in different <see cref="Team"/>s
+    /// in different <see cref="Database.Entities.Game"/>s
+    /// </summary>
+    RostersNotUnique = 1 << 3,
```
SourMongoose, line 37: Not sure if "unique" is the right way to describe what is being checked here - AB and AC are still unique rosters. Would be more accurate to use a name like `RostersNotMutuallyExclusive` or `OverlappingRosters`

majorro: agree, but let's see what @hburn7 thinks

DataWorkerService/AutomationChecks/Games/GameScoreCountCheck.cs
majorro: good catch, accidentally deleted team count check from the prev version, also added test on this

Common/Enums/Enums/Verification/MatchWarningFlags.cs
hburn7: I like "RostersNotMutuallyExclusive"

hburn7: Also please don't mark the thread as resolved if it's still awaiting feedback

hburn7: I'm also fine with overlapping

majorro: > Also please don't mark the thread as resolved if it's still awaiting feedback

yeah, sorry, probably I closed some extra

majorro: > I'm also fine with overlapping

lets go with it then

DataWorkerService/AutomationChecks/Matches/MatchTeamsIntegrityCheck.cs
```diff
@@ -0,0 +1,42 @@
...
+        foreach (Game game in validGames)
+        {
+            IEnumerable<GameScore> validScores = game.Scores.Where(x => x.VerificationStatus.IsPreVerifiedOrVerified());
+            game.Rosters = RostersHelper.GenerateRosters(validScores);
+        }
+
+        ICollection<MatchRoster> rosters = RostersHelper.GenerateRosters(validGames);
+        HashSet<int>[] playerIdsPerRoster = rosters.Select(x => x.Roster.ToHashSet()).ToArray();
+
+        for (var i = 0; i < playerIdsPerRoster.Length; ++i)
```
hburn7, line 28: Is there a reason for `++i` instead of `i++`? `++i` increments and then returns the state after it's incremented, so `i` will be equal to `1` on the first loop here.

hburn7 (changes requested):

DataWorkerService/AutomationChecks/Matches/MatchTeamsIntegrityCheck.cs
majorro: > Is there a reason for `++i` instead of `i++`

nope, just a habit

> `++i` increments and then returns the state after it's incremented, so` i` will be equal to `1` on the first loop here

no, it doesn't change anything in this case because the third statement is called after body finishing its execution

I thought about a rule that would automate it but haven't found it :( okay, I won't write it this way anymore 👌

SourMongoose: Don't think the return value matters for the `for` loop parameter, so `++i` here is just preference
I assume `i++` would be consistent with the rest of the code base though

hburn7: If there's no difference, then use `i++` please

Common/Enums/Enums/Verification/MatchWarningFlags.cs
```diff
@@ -28,5 +28,11 @@ public enum MatchWarningFlags
     /// The <see cref="Match"/> has 1 or more <see cref="Game"/>s with a <see cref="GameRejectionReason"/>
     /// of <see cref="GameRejectionReason.BeatmapNotPooled"/> outside of the first two <see cref="Game"/>s
     /// </summary>
-    UnexpectedBeatmapsFound = 1 << 2
+    UnexpectedBeatmapsFound = 1 << 2,
+
+    /// <summary>
+    /// The <see cref="Database.Entities.Match"/> has 1 or more <see cref="Database.Entities.Player"/>s that played in different <see cref="Team"/>s
```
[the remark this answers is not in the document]
majorro: 👌

DataWorkerService/AutomationChecks/Matches/MatchTeamsIntegrityCheck.cs
```diff
@@ -0,0 +1,42 @@
...
+/// <summary>
+/// Checks for <see cref="Match"/>es with the same <see cref="Player"/> among <see cref="Team"/>s
+/// </summary>
+public class MatchTeamsIntegrityCheck(ILogger<MatchTeamsIntegrityCheck> logger) : AutomationCheckBase<Match>(logger)
+{
+    public override int Order => 2;
+
+    protected override bool OnChecking(Match entity)
+    {
+        Game[] validGames = entity.Games.Where(x => x.VerificationStatus.IsPreVerifiedOrVerified()).ToArray();
```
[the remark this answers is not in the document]
majorro: Okay, i just like when i don't have to think what letter it should be 😄

DataWorkerService/AutomationChecks/Matches/MatchTeamsIntegrityCheck.cs
majorro: 👌

hburn7 (approved):
Should be good to go now

DataWorkerService.Tests/AutomationChecks/Matches/MatchTeamsIntegrityCheckTests.cs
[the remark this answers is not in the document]
majorro: Yeah, i thought the same, but i haven't seen similar extractions in other tests so far. Actually, the tests should be rewritten and extended imo, because this randomization is basically makes all of the tests prone to be flaky(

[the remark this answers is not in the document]
majorro: On this exact test - it isnt really copy-pasted, because it's just various match configurations written by hand
And i feel like i overengineered it instead of refactoring :p

PR #623, opened 2025-03-11
Disable forced collection expressions

majorro:
Disables [IDE0305](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0305) in editorconfig


majorro:
> Does `dotnet-format` pick up on this?

yes, if you see these warnings - it's just rider, I suppose
image

.editorconfig
```diff
@@ -106,6 +106,7 @@ dotnet_style_prefer_auto_properties = true:suggestion
 dotnet_style_prefer_conditional_expression_over_assignment = true:silent
 dotnet_style_prefer_conditional_expression_over_return = true:silent
 csharp_prefer_simple_default_expression = true:suggestion
+dotnet_style_prefer_collection_expression = false
```
hburn7, line 109:
```suggestion
dotnet_style_prefer_collection_expression = false:suggestion
```

majorro: it is disabled, why suggestion would be needed?

hburn7: It configures how the IDE reacts to this occurring in the project

majorro: yes, but since it's disabled+one way it won't show anything anywhere even if put `:error` here

but for overall consistency that's better ig

PR #627, opened 2025-03-14
Add thread safety for external API fetch operations

hburn7:
- Adds thread safety to DefaultRequestHandler by using a SemaphoreSlim with a limit of 1 concurrent thread. This prevents the osu! and osu!track APIs from being called by multiple threads simultaneously.
- Tournaments with a ProcessingStatus of NeedsMatchData will now execute one by one instead of concurrently. This operation was never thread safe.


myssto (commented):
lgtm

myssto (approved):
yep

PR #630, opened 2025-03-16
Refactor admin note routes into a single dynamic controller

myssto:


hburn7:
When testing via swagger, `gamescore` and `oauthclient` entities both return this. Is this intended or is swagger not using the correct entities in the dropdown?
image


```
{
  "errors": {
    "entity": [
      "The value 'gamescore' is not valid for DynamicRouteTarget."
    ]
  },
  "type": "https://tools.ietf.org/html/rfc9110#section-15.5.1",
  "title": "One or more validation errors occurred.",
  "status": 400,
  "traceId": "00-d2ae5dbac25bb8e8dd3e06a9bc624f6a-cfd3a2c14efc3a70-01"
}
```

API/Controllers/AdminNotesController.cs
```diff
@@ -0,0 +1,169 @@
...
+
+    /// <summary>
+    /// Delete an admin note
+    /// </summary>
+    /// <param name="noteId">Admin note id</param>
+    /// <response code="404">An admin note matching the given noteId does not exist </response>
+    /// <response code="400">The deletion was not successful</response>
+    /// <response code="204">The admin note was deleted</response>
+    // [Authorize(Roles = OtrClaims.Roles.Admin)]
+    [AllowAnonymous]
```
myssto, line 151:
```suggestion
    [Authorize(Roles = OtrClaims.Roles.Admin)]
```
Missed this from testing :P

myssto:
> When testing via swagger, `gamescore` and `oauthclient` entities both return this. Is this intended or is swagger not using the correct entities in the dropdown?

Nope this was a bug, good catch! The reason why some were working and others weren't was because originally I was using camel case to generate the route segments, but then switched to complete lowercase. This wasn't updated in the action filter so it was trying to match `oauthclient` to `oAuthClient` and `gamescore` to `gameScore`. Fixed by adding an extension to share the formatted string output between the swagger doc and the action filter

API/Controllers/AdminNotesController.cs
```diff
@@ -0,0 +1,168 @@
...
+ *
+ * These inspections are failing because the {entity} segment of the RouteAttribute defined
+ * on the controller itself is not directly passed to each method. Instead of being passed to
+ * each method, it is bound to a property of the class, which is implicitly accessible to each method.
+ */
+
+[ApiController]
+[ApiVersion(1)]
+[ValidateAdminNoteControllerRoute]
+[Route("api/v{version:apiVersion}/{entity}")]
```
hburn7, line 34: Is there a reason why the entity is forced at this level? For endpoints that only rely on the primary key of the note directly, such as deletion, it shouldn't matter what the entity is.

hburn7 (approved):

PR #632, opened 2025-03-18
Replace semaphore with global mutex

hburn7:
Adds support for fail-safe rate limiting in the event of a 429 response. This fixes an issue where the client will continue to spam if it thinks it has tokens available when the platform returns 429. Also moves the semaphore to surround the entire request lifecycle rather than just one part of it. This includes the part where rate limits are being respected. My assumption is that previously, some parts of the request lifecycle were thread safe while others weren't.

image

### DIFF
--- a/OsuApiClient/Net/Requests/RequestHandler/DefaultRequestHandler.cs
+++ b/OsuApiClient/Net/Requests/RequestHandler/DefaultRequestHandler.cs
@@ -1,3 +1,4 @@
+using System.Net;
 using System.Net.Http.Headers;
 using System.Text;
 using AutoMapper;
@@ -104,13 +105,17 @@ internal sealed class DefaultRequestHandler(
         CancellationToken cancellationToken = default
     )
     {
-        HttpRequestMessage requestMessage = await PrepareRequestAsync(request, cancellationToken);
-        HttpResponseMessage? response = null;
-
         try
         {
             await _semaphore.WaitAsync(cancellationToken);
-            response = await _httpClient.SendAsync(requestMessage, cancellationToken);
+
+            HttpRequestMessage requestMessage = await PrepareRequestAsync(request, cancellationToken);
+            HttpResponseMessage response = await _httpClient.SendAsync(requestMessage, cancellationToken);
+
+            var responseContent = await OnResponseReceivedAsync(request.Platform, response, cancellationToken);
+
+            _semaphore.Release();
+            return responseContent;
         }
         catch (HttpRequestException ex)
         {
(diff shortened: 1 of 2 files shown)


myssto:
Unfortunately I think this problem is a bit harder to fix than this. The semaphore should prevent multiple requests from being sent at once, but that is only going to apply for that instance of the request handler. In our stack we have multiple projects (currently API and DataWorkerService) each registering its own instance of the client. Because of this there is still a chance that a request is made to the osu! API from our domain while one is awaiting completion (for example a user requests authorization with the API while the DataWorkerService is in the middle of a data fetching task). This is just my initial thought and I don't have a plan to fix but I think this is worth looking into a bit more in depth so we can put it to bed for good instead of slapping a band-aid on it and calling it a day

hburn7:
Updated to use named mutex.

OsuApiClient/Net/Requests/RequestHandler/DefaultRequestHandler.cs
```diff
@@ -21,7 +22,8 @@ internal sealed class DefaultRequestHandler(
     IOsuClientConfiguration configuration
 ) : IRequestHandler
 {
-    private readonly SemaphoreSlim _semaphore = new(1, 1);
+    private const string GlobalMutexName = "OTR_2b1c2e2f-9c62-4433-9066-dfb7a397d03d";
```
myssto, line 25:
```suggestion
    private const string GlobalMutexName = "Global\OTR_2b1c2e2f-9c62-4433-9066-dfb7a397d03d";
```

https://learn.microsoft.com/en-us/dotnet/api/system.threading.mutex.-ctor?view=net-9.0#system-threading-mutex-ctor(system-boolean-system-string)

> The name may be prefixed with Global\ or Local\ to specify a namespace. When the Global namespace is specified, the synchronization object may be shared with any processes on the system. When the Local namespace is specified, which is also the default when no namespace is specified, the synchronization object may be shared with processes in the same session.

I guess we need to use this convention for the name

OsuApiClient/Net/Requests/RequestHandler/DefaultRequestHandler.cs
```diff
@@ -96,21 +98,33 @@ public async Task FetchAsync(
     /// <summary>
     /// Sends a request
     /// </summary>
+    /// <remarks>
+    /// Everything inside of this method is controlled by a named <see cref="Mutex"/>.
+    /// Each operation must execute on the same thread, therefore asynchronous calls are not possible.
+    /// The Mutex exists to prevent the situation where multiple processes are using the same
+    /// osu! API client and calling multiple requests against it simultaneously.
+    /// </remarks>
```
myssto, line 106:
```suggestion
    /// <remarks>
    /// Execution of this method is protected by a named <see cref="Mutex"/>.
    /// Each operation must execute on the same thread, therefore asynchronous calls are not possible.
    /// The mutex exists to prevent situations where multiple processes utilizing the request handler
    /// attempt to make network requests at the same time
    /// </remarks>
```

myssto (changes requested):

PR #635, opened 2025-03-23
Add distributed locking support for osu! requests

hburn7:
Replaces the (failed) mutex implementation with a distributed lock implemenation. This should be safe across isolated containers as the lock is managed by redis. Updates the example appsettings files to reflect the ideal defaults for fetching.

This implementation works and was tested locally using /oauth/authorize with a dummy value while the DataWorkerService was processing (both osu! and osu!track API calls). Our `/oauth/authorize?code=...` endpoint does compete with the DataWorkerService for osu! API requests. However, with a small enough batch size and a generous rate limit, we shouldn't have many issues. If the rate limit is fully consumed, requests will wait until the rate limit becomes available again (this includes requests made to `/oauth/authorize?code=...`).

This fixes a bug where, on production, users cannot login while the DataWorkerService is running. This (hopefully) fixes another related bug where the DataWorkerService will encounter 429 errors while fetching data from the osu! API.


hburn7:
Blocking until I update the staging and prod DataWorkerService config with the new RedisConnection string.

myssto (approved):

myssto:
dont you need to edit dockerfile(s) to make data worker depend on redis?

hburn7:
Funnily the dataworkerservice already depends on redis. It was meant to be.

hburn7:
Updated the configs

PR #640, opened 2025-03-30
Include `User` navigations in `AdminNote`s

hburn7:
Fixes an issue where some required chain navigations were not included in `AdminNote` DTOs.


hburn7:
I did notice a bit of performance degradation, takes a few seconds to load the data from the `/matches/id` endpoint. I'll investigate before merging.

myssto (approved):

PR #642, opened 2025-04-01
Fix tier logic in player stats output

hburn7:
## Dependencies

- [x] #641 

The previous tier logic didn't make a whole lot of sense. We were displaying strings with roman numerals *and* a "subtier" field. In reality, this display should be handled by the caller.

## Overivew

This change:

- Removes a duplicate field from the stats output (currentTier was present in rating and rankProgress).
- Removes the string display of sub tiers from the stats endpoint object
- Uses display of "major tiers" where we used to display the full tier with roman numerals.

image

### DIFF
--- a/API.Tests/Utilities/RatingUtilsTests.cs
+++ b/API.Tests/Utilities/RatingUtilsTests.cs
@@ -6,37 +6,37 @@ namespace APITests.Utilities;
 public class RatingUtilsTests
 {
     [Theory]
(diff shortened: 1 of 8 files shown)


SourMongoose:
feel like it would make more sense to include `"currentTier"` within the `"rankProgress"` dict

myssto (approved):

hburn7:
Renamed RankProgress to TierProgress, Tier is definitely the correct prefix.

PR #648, opened 2025-04-22
Remove default rate limit config values and anonymous limit partition

hburn7:
- Removes default rate limit configuration values. These values should always be explicitly set in the configuration.
- Removes anonymous shared rate limit bucket. Cloudflare will handle spammers at the network level. This also is a bad idea to have because if we have a high user count, the service will be unbearably rate limited by innocent users.

### DIFF
--- a/API/Program.cs
+++ b/API/Program.cs
@@ -187,15 +187,6 @@ builder.Services.AddRateLimiter(options =>
     // Configure the rate limit partitioning rules
     options.GlobalLimiter = PartitionedRateLimiter.Create<HttpContext, string>(context =>
     {
-        // Shared partition for anonymous requests
-        if (context.User.Identity is { IsAuthenticated: false })
-        {
-            return RateLimitPartition.GetFixedWindowLimiter(
-                "anonymous",
-                _ => GetRateLimiterOptions()
-            );
-        }
-
         // Partition for each unique user / client
         return RateLimitPartition.GetFixedWindowLimiter(
             $"{context.User.GetSubjectType()}_{context.User.GetSubjectId()}",
--- a/API/Configurations/RateLimitConfiguration.cs
+++ b/API/Configurations/RateLimitConfiguration.cs
@@ -7,13 +7,14 @@ public class RateLimitConfiguration
     public const string Position = "RateLimit";
 
     /// <summary>
-    /// The total amount of tokens available per <see cref="Window"/>
+    /// The total amount of tokens available per <see cref="Window"/>, per user/client
     /// </summary>
     [Range(1, int.MaxValue, ErrorMessage = "PermitLimit must be an integer greater than 1!")]
-    public int PermitLimit { get; init; } = 30;
+    public int PermitLimit { get; init; }
+
     /// <summary>
     /// The amount of time (in seconds) before the <see cref="PermitLimit"/> is refreshed
     /// </summary>
     [Range(1, int.MaxValue, ErrorMessage = "Window must be an integer greater than 1!")]
-    public int Window { get; init; } = 60;
+    public int Window { get; init; }
 }


API/Program.cs
```diff
@@ -187,15 +187,6 @@ await context.HttpContext.Response.WriteAsync(
...
     {
-        // Shared partition for anonymous requests
-        if (context.User.Identity is { IsAuthenticated: false })
-        {
-            return RateLimitPartition.GetFixedWindowLimiter(
-                "anonymous",
-                _ => GetRateLimiterOptions()
-            );
-        }
-
```
myssto, line 198: Besides the fact that I don't even agree with this change, this is unsafe to remove. The following code after this section:

```cs
        return RateLimitPartition.GetFixedWindowLimiter(
            $"{context.User.GetSubjectType()}_{context.User.GetSubjectId()}",
            _ => GetRateLimiterOptions(context.User.GetRateLimitOverride())
        );
```

Will throw if the user is not authenticated because of the use of `context.User.GetSubjectId()`.

myssto (changes requested):
> Removes default rate limit configuration values. These values should always be explicitly set in the configuration.

I don't see the reason for this at all. All this change would do is force a developer to supply more values in configuration that they don't care about.

API/Program.cs
hburn7: Is there a reason why you don't agree with this specific change? My goal with the change is to remove the anonymous bucket (effectively making it unlimited), because CloudFlare will handle this and a vast majority of our routes require auth to use. The data we serve without auth is well-cached anyway isn't it?

This is what I would like to do:

```cs
// Configure the rate limit partitioning rules
    options.GlobalLimiter = PartitionedRateLimiter.Create<HttpContext, string>(context =>
    {
        if (context.User.Identity is null or { IsAuthenticated: false })
        {
            return RateLimitPartition.GetNoLimiter("anonymous");
        }

        // Partition for each unique user / client
        return RateLimitPartition.GetFixedWindowLimiter(
            $"{context.User.GetSubjectType()}_{context.User.GetSubjectId()}",
            _ => GetRateLimiterOptions(context.User.GetRateLimitOverride())
        );
    });
```

hburn7:
I think the only context in which it makes sense to allow defaults is when running through docker. In prod, we specify a full config file. To be compliant with TC, nobody will need to run this through docker. Thus it can be assumed that anyone running the API is a developer. In my opinion, I think those using our application need to be aware of what the cfg fields are and what they're doing.

I don't think it's a good idea to be setting rate limits for people on their behalf by default. I think the sensible default is to force those running this app to be aware of what the configuration fields are and provide a template for them to use. I don't want to bear the responsibility of setting a default rate limit targeting an external service, if I were the party receiving that traffic I don't think I'd appreciate it.

myssto:
> I don't think it's a good idea to be setting rate limits for people on their behalf by default. I think the sensible default is to force those running this app to be aware of what the configuration fields are and provide a template for them to use. I don't want to bear the responsibility of setting a default rate limit targeting an external service, if I were the party receiving that traffic I don't think I'd appreciate it.

Also huge ? to this... We are not obligated to do jack for any third party wanting to run our app. It's a heavily documented section of configuration, it's not like someone running the API can't just change it with config? And if said third party can't be bothered to read the documentation then I wouldn't care in the first place what they wanted to do.

API/Program.cs
myssto: > My goal with the change is to remove the anonymous bucket (effectively making it unlimited), because CloudFlare will handle this

I understand that is what the goal is, I just have no idea what "Cloudflare will handle it" is supposed to mean tbh. By that line of logic would it not make more sense to just remove the rate limiter for users entirely and only limit clients? I just don't see the harm in this being here and it's good presence of mind to know something baked into our service will have our back in case "Cloudflare will handle it" is suddenly not the case anymore.

> The data we serve without auth is well-cached anyway isn't it?

Don't think we currently serve anything without auth, and nothing we serve is cached currently

hburn7: Cloudflare will block access at the network level if the same IP spams any /api/v1* request too many times. This rate limit feature is [pretty restrictive](https://developers.cloudflare.com/waf/rate-limiting-rules/) without paying for more features, but it should be good enough for us. This is beneficial because it blocks the traffic before it reaches our server, saving compute on our end.

Keeping the rate limiter at the user level is still good for us because we can mitigate spam this way too, saving compute. Cloudflare's block is basically the worst case scenario regardless of user or client permissions. So our rate limit will be more granular (and less permissive) than what Cloudflare offers.

The problem with a shared anonymous bucket is that it doesn't scale. If we have spikes in traffic, 429s are going to pop up everywhere, even for users who are not doing anything wrong. I'd rather remove this and see if we run into significant issues with this kind of traffic first. We still don't know what our server is capable of. Most websites will heavily limit functionality to users who aren't logged in. Search, downloads, and other requests which can't be easily cached or require heavy compute to use should be limited to authorized users only. In our case, I think displaying player pages, leaderboards, and browsing tournaments should be okay for anonymous usage (from otr-web specifically). This will probably require us to introduce a shared secret between web and the API so that we know certain anonymous requests are coming from web. I don't think we should allow any truly anonymous API requests that are not originated from web.

hburn7:
After some online reading, I can get behind the sensible defaults. I would like to introduce some logging functionality which will print out the state of non-secret configuration values so that we know for sure things are being read in as we expect. The root of my concern is the situation where cfg fields are unexpectedly falling back to the default.

API/Program.cs
```diff
@@ -187,13 +187,12 @@ await context.HttpContext.Response.WriteAsync(
     // Configure the rate limit partitioning rules
     options.GlobalLimiter = PartitionedRateLimiter.Create<HttpContext, string>(context =>
     {
-        // Shared partition for anonymous requests
-        if (context.User.Identity is { IsAuthenticated: false })
+        // Unlimited partition for anonymous requests, as doing so
+        // would cripple the ability for the platform to support
+        // an influx of legitimate anonymous requests
+        if (context.User.Identity is null or { IsAuthenticated: false })
```
myssto, line 193:
```suggestion
        if (context.User.Identity is { IsAuthenticated: false })
```

null check here is not necessary, the pattern check will fail if it is null

myssto (approved):
Just that and one change and then is fine

PR #659, opened 2025-04-29
Code quality refactor of PlayerService and FilteringService

### LINKED ISSUE #305: Code quality refactor of the `PlayerService`
Like the `PlayerRepository`, the `PlayerService` needs a bit of cleanup as well. Because many of the tasks are shared with 304, I will create a task list for this issue once that has been completed.

Depends on #304

hburn7:
Focused on quality, not efficiency. Filtering functionality still works great.

Closes #305


API/Services/Implementations/FilteringService.cs
```diff
@@ -12,113 +12,118 @@ public class FilteringService(
...
+
+        return new PlayerFilteringResultDTO
+        {
+            PlayerId = playerInfo?.Id,
+            Username = playerInfo?.Username,
+            OsuId = playerInfo?.OsuId ?? 0,
+            FilteringResult = result,
+            FilteringFailReason = failReason
         };
     }
```
myssto, line 60: don't understand why this was moved into it's own function when it is only used once. would rather this just be inlined where it was before

API/Services/Implementations/FilteringService.cs
```diff
@@ -12,113 +12,118 @@ public class FilteringService(
...
         }
 
+        // Count results and create result DTO
+        var passed = results.Count(r => r.FilteringResult == FilteringResult.Pass);
+        var failed = results.Count(r => r.FilteringResult == FilteringResult.Fail);
+
         return new FilteringResultDTO
         {
             PlayersPassed = passed,
             PlayersFailed = failed,
```
myssto, line 41:
```suggestion
        return new FilteringResultDTO
        {
            PlayersPassed = results.Count(r => r.FilteringResult == FilteringResult.Pass),
            PlayersFailed = results.Count(r => r.FilteringResult == FilteringResult.Fail),
```

API/Services/Implementations/FilteringService.cs
```diff
@@ -12,113 +12,118 @@ public class FilteringService(
...
     public async Task<FilteringResultDTO> FilterAsync(FilteringRequestDTO filteringRequest)
     {
+        ArgumentNullException.ThrowIfNull(filteringRequest);
+
         var idList = filteringRequest.OsuPlayerIds.ToList();
 
         if (idList.Count == 0)
         {
-            throw new Exception("Filtering id list cannot be empty");
+            throw new ArgumentException("Filtering id list cannot be empty", nameof(filteringRequest));
```
myssto, line 21: don't see the need for this to throw considering none of the operations depending on this list will cause an unrecoverable error if it is empty. this should probably be validated at the controller, or better yet with a data annotation on the request body which cause it the request to fail at validation time

API/Services/Implementations/FilteringService.cs
```diff
@@ -12,113 +12,118 @@ public class FilteringService(
 {
     public async Task<FilteringResultDTO> FilterAsync(FilteringRequestDTO filteringRequest)
     {
+        ArgumentNullException.ThrowIfNull(filteringRequest);
```
myssto, line 15: when will this ever have a chance of being null? I understand this is common practice in c# but we never do this anywhere in the codebase

API/Services/Implementations/FilteringService.cs
```diff
@@ -12,113 +12,118 @@ public class FilteringService(
...
 
         if (idList.Count == 0)
         {
-            throw new Exception("Filtering id list cannot be empty");
+            throw new ArgumentException("Filtering id list cannot be empty", nameof(filteringRequest));
         }
 
-        var passed = 0;
-        var failed = 0;
+        IEnumerable<PlayerCompactDTO?> players = await playerService.GetAsync(idList);
```
myssto, line 24: don't really need to define `idList` as it's own variable, just pass it to this function like `playerService.GetAsync(filteringRequest.OsuPlayerIds)`

API/Services/Implementations/FilteringService.cs
```diff
@@ -12,113 +12,118 @@ public class FilteringService(
...
+        FilteringFailReason failReason = CheckFilteringConditions(request, ratingStats);
+
+        var peakRating = await playerStatsService.GetPeakRatingAsync(playerInfo.Id, request.Ruleset);
+        if (peakRating > request.PeakRating)
+        {
+            failReason |= FilteringFailReason.PeakRatingTooHigh;
         }
 
-        if (ratingStats.Rating < filteringRequest.MinRating)
+        FilteringResult result = failReason == FilteringFailReason.None ? FilteringResult.Pass : FilteringResult.Fail;
```
myssto, line 90: wait so why do even have an enum `FilteringResult` for this when it literally represents a boolean `pass` or `fail` LOL? This function should just return a `FilteringFailReason` and whether or not filtering failed should be implicit based on the flags

API/Services/Implementations/FilteringService.cs
```diff
@@ -12,113 +12,118 @@ public class FilteringService(
...
+
+        return (result, failReason);
+    }
```
myssto, line 93: Why are this function and `CheckFilteringConditions` separate? Why does `CheckFilteringConditions` only return a `FilteringFailReason` and not a tuple like this one does (which it shouldn't in the first place)?

myssto (changes requested):
Will review rest after changes

API/Services/Implementations/FilteringService.cs
hburn7: Eh yeah I shouldn't have put this here since it's not a nullable type, will remove.

hburn7:
- Refactored a lot of the AI jank
- Created a new `SupportedRuleset` validation attribute to handle the case where `osu!mania (Other)` is provided. I think we could use this as a data annotation on all of our DTOs which have a Ruleset field.
- Replaced the `FilteringResult` enum with a boolean that switches based on the state of the failure reason.

API/Services/Implementations/PlayerService.cs
```diff
@@ -8,45 +8,24 @@ namespace API.Services.Implementations;
...
+            .Where(p => p is not null)
+            .ToDictionary(p => p!.OsuId);  // Assuming Player has OsuId property
+
+        // Return a list matching the input order:
+        // - If the player exists in our dictionary, return the mapped DTO
+        // - If not found, return null
+        return idList.Select(id =>
+            players.TryGetValue(id, out Player? player)
+                ? mapper.Map<PlayerCompactDTO>(player)
+                : null);
```
myssto, line 29: Is there a reason that this very specific return format is done here instead of where the data is actually being used? seems a bit strange considering this is probably only used in one place right?

myssto: to be more specific, I don't necessarily see why it should be the responsibility of this service to provide a null in place when a player is not found out of a list of ids. its not consistent with any other impl of `IEnumerable<Entity> GetAsync(IEnumerable<int> ids)` in the codebase. caller of the function should be able to assume that if no player was returned for an id that it input, one does not exist.

API/DTOs/FilteringRequestDTO.cs
```diff
@@ -59,5 +60,6 @@ public class FilteringRequestDTO
     /// <summary>
     /// A list of osu! player ids that will be filtered
     /// </summary>
-    public required IEnumerable<long> OsuPlayerIds { get; set; } = [];
+    [MinLength(1, ErrorMessage = "At least one osu! player ID is required.")]
```
myssto, line 63:
```suggestion
    [MinLength(1)]
```
I have tried to avoid using custom error messages if at all possible

API/DTOs/PlayerFilteringResultDTO.cs
```diff
@@ -1,38 +1,51 @@
...
-    /// <summary>
-    /// The <see cref="FilteringResult"/> in string form
-    /// </summary>
-    public string FilteringResultMessage => FilteringResult.ToString();
+
     /// <summary>
     /// The <see cref="FilteringFailReason"/> in string form
     /// </summary>
-    public string? FilteringFailReasonMessage => FilteringFailReason.ToString();
+    public string? FilteringFailReasonMessage => FilteringFailReason is null ? null : FilteringFailReasonMessage;
```
myssto, line 50: do we even really need this? we use raw enums in every other DTO, don't see how this one should be treated different

API/DTOs/PlayerFilteringResultDTO.cs
```diff
@@ -1,38 +1,51 @@
...
     /// <summary>
     /// The osu! id of the player
     /// </summary>
     public long OsuId { get; set; }
+
     /// <summary>
-    /// The filtering result
+    /// Whether the player successfully passed through the filter.
+    /// If this value is true, the player meets all constraints
+    /// specified in the filter.
```
myssto, line 30:
```suggestion
    /// Whether the player successfully passes all conditions of the filter
```

API/DTOs/PlayerFilteringResultDTO.cs
```diff
@@ -1,38 +1,51 @@
...
+    /// whose peak rating is less than 1000.
+    ///
+    /// If Player A's rating is 800 and Player B's
+    /// rating is 1001, this value will be true
+    /// for Player A and false for Player B.
+    /// </example>
+    public bool IsSuccess { get; set; }
+
     /// <summary>
     /// If the user failed filtering, the fail reason
```
myssto, line 43:
```suggestion
    /// Reason for failure if applicable
```

API/Services/Implementations/FilteringService.cs
```diff
@@ -10,115 +10,104 @@ public class FilteringService(
...
-                PlayerId = playerInfo.Id,
-                Username = playerInfo.Username,
-                OsuId = playerInfo.OsuId,
-                FilteringResult = result,
+                PlayerId = playerInfo?.Id,
+                Username = playerInfo?.Username,
+                OsuId = playerInfo?.OsuId ?? 0,
                 FilteringFailReason = failReason
             });
         }
```
myssto, line 29: this could probably be done nicely with a select statement

API/Services/Implementations/FilteringService.cs
```diff
@@ -10,115 +10,104 @@ public class FilteringService(
...
-        if (failReason == FilteringFailReason.None)
+        if (ratingStats.MatchesPlayed < request.MatchesPlayed)
         {
-            result = FilteringResult.Pass;
+            failReason |= FilteringFailReason.NotEnoughMatches;
         }
 
-        return (result, failReason);
+        return failReason;
     }
```
myssto, line 112: these two functions should just be merged

Common/Enums/FilteringFailReason.cs
```diff
@@ -1,43 +1,51 @@
...
 /// <summary>
-/// Explains why the player failed filtering
+/// Indicates why the player failed filtering
 /// </summary>
 [Flags]
 public enum FilteringFailReason
 {
     /// <summary>
-    /// The player passed filtering, thus there is no failure reason
+    /// The player does not exist in the system at all
```
myssto, line 10:
```suggestion
    /// The player is not indexed by o!TR
```
reads a bit nicer maybe?

myssto (changes requested):

API/Services/Implementations/FilteringService.cs
hburn7: I originally tried with a select statement but it ends up returning a task list. Even with an async lambda, I can't await each filter result immediately.

API/Services/Implementations/FilteringService.cs
hburn7: I disagree, I think these should be split because this is the single location where all filtering enforcement will live (separation of concerns). This is split on the same grounds as `FilterPlayerAsync`'s logic being separated from `FilterAsync`.

API/Services/Implementations/PlayerService.cs
hburn7: I was mainly trying to avoid looping through the list a second time to determine whether anyone is missing. In practice that is kinda jank and that's what a Dictionary is for. Even with a Dictionary impl, another iteration would need to be done somewhere to identify users who weren't fetched.

I think it's important to not leave room for assumptions in this case (from the perspective of the API caller not knowing why there are missing results). We can include a list of IDs which returned no result instead of including nulls in the output. I'll redo this impl with that in mind.

API/Services/Implementations/FilteringService.cs
myssto: dont really have a strong opinion either way

API/Services/Implementations/FilteringService.cs
myssto: that's true, didn't think of that and they aren't safe to multithread because of db interface

myssto:
> I was mainly trying to avoid looping through the list a second time to determine whether anyone is missing. In practice that is kinda jank and that's what a Dictionary is for. Even with a Dictionary impl, another iteration would need to be done somewhere to identify users who weren't fetched.

I understand and respect wanting to preemptively optimize but to be frank I seriously think it just doesn't matter. I would rather things be a bit easier to follow and work with than creating incredibly specific single-use methods on a generic repository. Imo doing it the way it is currently with the `NotFoundOsuIds` list in the response is silly and doesn't align with anything else in the codebase. And fwiw, `NotFoundOsuIds = [.. osuIdHashSet.Except(players.Select(p => p.OsuId))]` is going to cause that extra enumeration to occur that you were trying to avoid in the first place. At that point why not just take the resulting difference and map them to a result anyway?

 ```cs
osuIdHashSet
  .Except(players.Select(p => p.OsuId))
  .Select(osuId => {
      return new PlayerFilteringResultDTO
      {
          OsuId = osuId,
          IsSuccess = false,
          FailureReason = FilteringFailReason.NoData
      }
  })
```

I think you should just:
- Receive `IEnumberable<long> osuIds` from filter request
- Get `IEnumerable<Player> players` from database
- Iterate over `osuIds`, **not** `players`
- Determine if player is missing with `players.FirstOrDefault(p => p.OsuId == osuId)`

tada every osu id in the request gets a result

PR #668, opened 2025-05-11
Add endpoint to merge matches

### LINKED ISSUE #616: Add ability to merge matches
# Overview

This feature should combine the games from two different MP links into a single match. Specifically, the games from the second match should have their parent match ID reassigned to that of the first match (leaving the second match with no games), and the games should still show up in the correct temporal order.

After this, verification / processing status for the first match should probably revert to None / NeedsAutomationChecks; ditto for the second match if it is not automatically deleted. The operation should also be logged in the same location as other "manual edits to match data."

# Dependencies
~Probably a web dependency rather than an API one, but there should be admin note functionality at the match level (osu-tournament-rating/otr-web#55) so that the merge can be publicly documented.~

API support for admin notes is complete

# Tasks
- [ ] Create a new endpoint
  - Method `POST`
  -  Route `/matches/[id]:transfer-games?target=[second_id]`
  - Roles `admin`
  - Action: For each game in the match with id of `second_id`, set the `game.match_id` value to `id`.

hburn7:
Closes #616


Database/Repositories/Implementations/MatchesRepository.cs
```diff
@@ -115,6 +116,35 @@ public async Task<IEnumerable<Match>> SearchAsync(string name)
...
+
+        ICollection<Match> childMatches = await _context.Matches
+            .Include(m => m.Games)
+            .Where(m => matchIds.Contains(m.Id))
+            .ToListAsync();
+
+        childMatches.ForEach(child => child.Games.ForEach(childGame =>
+        {
+            childGame.Match = parentMatch;
+            MarkUpdated(childGame);
```
myssto, line 136: not ever necessary to do this, see `Database.OtrContext.cs:991`

API/Controllers/MatchesController.cs
```diff
@@ -45,6 +45,26 @@ public async Task<IActionResult> GetAsync(int id, [FromQuery] bool verified = tr
             : Ok(match);
     }
 
+    /// <summary>
+    /// Merge the games from provided matches into the parent match, then delete the provided matches.
+    /// </summary>
+    /// <param name="id">The parent match id</param>
+    /// <param name="matchIds">The match ids which games should be moved from before being deleted</param>
+    /// <response code="404">The parent match id could not be found</response>
+    /// <response code="200">The newly updated match after the merge operation</response>
```
myssto, line 54: please conform to the verbiage of the rest of the endpoint documentation

myssto (changes requested):

Database/Repositories/Implementations/MatchesRepository.cs
hburn7: Nice, didn't know that was a feature

Issue #650, opened 2025-04-27
Transfer session management from web to API


hburn7:
# Overview

We want to transfer session management from the website to the API. The frontend currently uses NextAuth, backend currently does not have anything to do with session management. The end goal is to use cookie-based authentication on the API which will then result in sessions being stored as a cookie.

We are making this change because the setup we're currently using is highly customized and is quite messy. As we have a backend for frontend architecture, it makes sense to leverage the backend (which reaches out to osu! OAuth) such that it handles session management.

# Tasks

- [x] Add cookie-based authentication to API
- [x] Remove `OAuthController` and move any functionality that we need to keep for API clients into the `ClientsController`.

PR #673, opened 2025-05-12
Add support for cookie-based authentication

myssto:
Part of #650

Adds the ability for clients to authenticate via a cookie managed by ASP. This authentication method is implemented on top of the existing JWT Bearer authentication, and both methods serialize to a valid `ClaimsPrincipal`. In practice, this should enable the API to handle persistent user authentication, while still allowing API client access with bearer tokens. On top of this, user JWTs created by the JWT Tool will continue to function as normal for development purposes.

### Key Changes
- Adds cookie-based authentication
- Remove refresh tokens
- Remove most of the existing auth service chain (`OAuthHandler`, `OAuthController`, `JwtService`)
- Refactor login logic into a new `AuthController`
- Refactor API client token generation into the existing `OAuthClientService` and `ClientsController`

### Config Changes
- Adds ability to configure the valid CORS origins `Auth.AllowedHosts`
- Requires `{domain}/api/v1/auth/callback` be added as a redirect uri to the osu! OAuth client in use


myssto:
Blocked for massive amounts of breaking changes. Requires an update of `web` and `api-clients` to be compatible

API/Program.cs
```diff
@@ -494,32 +480,114 @@ await context.HttpContext.Response.WriteAsync(
...
+                    cancellationToken: context.HttpContext.RequestAborted
+                );
 
-                return Task.CompletedTask;
+                // Get or create necessary entities
+                Player player = await playersRepository.GetOrCreateAsync(osuUser!.Id);
+
+                User user = await usersRepository.GetByPlayerIdOrCreateAsync(player.Id);
+                user.LastLogin = DateTime.UtcNow;
+                await usersRepository.UpdateAsync(user);
```
hburn7, line 575: Could we extract this into something like `usersService.LoginAsync` that does this? We could also implement friends list syncing there in the future.

OsuApiClient/OsuClient.cs
```diff
@@ -47,6 +47,9 @@ public void Dispose()
         _disposed = true;
     }
 
+    public OsuCredentials SetCredentials(OsuCredentials credentials) =>
```
hburn7, line 50: I think you can use `public OsuCredentials? Credentials { private get; set; }` instead.

hburn7 (changes requested):
Looks good, just a few small things from me. Let's document (here or in dev chat) what changes we need to make on ops side to support this.

API/Program.cs
myssto: Sure we could, and probably a good idea. This function body was mostly copy pasted from the old oauth handler code

hburn7:
Should be ready to merge once web changes are done.

PR #675, opened 2025-05-13
Use `.AsSplitQuery` to improve `Match` query performance

hburn7:


Database/Repositories/Implementations/MatchesRepository.cs
```diff
@@ -95,6 +95,7 @@ public async Task<IEnumerable<Match>> GetAsync(IEnumerable<long> matchIds) =>
         IQueryable<Match> query = _context.Matches
             .AsSplitQuery()
             .AsNoTracking()
+            .AsSplitQuery()
```
majorro, line 98: why is this second call needed?

hburn7: Oh nvm I thought this wasn't even implemented in the first place

hburn7:
Already implemented whoops

PR #716, opened 2025-05-31
Fix cookie key persistence

### LINKED ISSUE #694: Cookie seemingly invalidates when server restarts
Needs investigation as to whether it's the API or web causing this. When one of the two applications restarts, the user needs to log out and log in again before the site functions properly. Ideally the cookie persists through restarts.

hburn7:
Closes #694


hburn7:
Didn't experience the restart glitch on staging so, worth a shot I guess

PR #736, opened 2025-06-13
Add support for merging games

hburn7:
Adds functionality which enables game merging. The "rules" of merging games are:

- All games must be played in the same beatmap in the same match
- All games must not have overlapping scores for a player. Admins should delete unwanted scores from child games before merging them into the parent game. If keeping the child scores, delete scores from the parent first.

### DIFF
--- a/API/API.csproj
+++ b/API/API.csproj
@@ -8,7 +8,6 @@
     <Nullable>enable</Nullable>
     <ImplicitUsings>enable</ImplicitUsings>
     <DockerDefaultTargetOS>Linux</DockerDefaultTargetOS>
-    <UserSecretsId>dad212ce-6224-4527-8e02-14264d5bf7e7</UserSecretsId>
   </PropertyGroup>
   <PropertyGroup>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
--- a/DataWorkerService/DataWorkerService.csproj
+++ b/DataWorkerService/DataWorkerService.csproj
@@ -4,7 +4,6 @@
     <TargetFramework>net9.0</TargetFramework>
     <Nullable>enable</Nullable>
     <ImplicitUsings>enable</ImplicitUsings>
-    <UserSecretsId>dotnet-DataWorkerService-DABD3AE0-FAD2-401E-B7F7-A2F931C01155</UserSecretsId>
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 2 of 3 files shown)


SourMongoose (approved):

Database/Repositories/Implementations/GamesRepository.cs
```diff
@@ -28,4 +29,105 @@ await _context.Entry(game)
...
+        try
+        {
+            Game? targetGame = await _context.Games
+                .Include(g => g.Match)
+                .Include(g => g.Scores)
+                .FirstOrDefaultAsync(g => g.Id == targetGameId);
+
+            if (targetGame is null)
+            {
+                logger.LogDebug("Game merge failed: Target game {TargetGameId} not found", targetGameId);
```
majorro, line 45: why all of these are debug?

Database/Utilities/Extensions/QueryExtensions.cs
```diff
@@ -404,7 +404,12 @@ public static IQueryable<Game> IncludeChildren(this IQueryable<Game> query, bool
         }
 
         return query
+            .Include(g => g.Match)
```
majorro, line 407: docs update is needed ig

Database/Repositories/Implementations/GamesRepository.cs
```diff
@@ -28,4 +29,105 @@ await _context.Entry(game)
...
+            {
+                score.Game = targetGame;
+            }
+
+            // Save before deleting source games
+            await _context.SaveChangesAsync();
+
+            // Delete source games using tracked deletion to trigger auditing
+            _context.Games.RemoveRange(sourceGames);
+            await _context.SaveChangesAsync();
```
majorro, line 118: i think this should be wrapped into transaction as it's a unit action being split into two saves

Database/Repositories/Implementations/GamesRepository.cs
```diff
@@ -28,4 +29,105 @@ await _context.Entry(game)
             .Collection(g => g.Scores)
             .LoadAsync();
     }
+
+    public async Task<Game?> MergeScoresAsync(int targetGameId, IEnumerable<int> sourceGameIds)
```
majorro, line 33: shouldn't match automation checks be reset here? and maybe restrict this merge action to some state/status for data integrity?

majorro, line 33: checks can be extracted to make this method shorter, but it's a matter of preference in this case

Database/Repositories/Implementations/GamesRepository.cs
hburn7: If an admin has a question about why the merge failed, we can go into the debug logs and see why. Not important enough to be an info log imo.

PR #754, opened 2025-06-14
Fix authentication for public endpoints

hburn7:
Added explicit [AllowAnonymous] attribute to endpoints that were recently made public by removing [Authorize] attributes. This fixes authentication scheme selection issues causing 401 errors on /leaderboards and /tournaments endpoints. The authentication middleware now properly handles both authenticated and anonymous users on these endpoints.


hburn7:
Tested on staging, resolves the crash on /leaderboard and failure to fetch /tournaments data

SourMongoose (approved):

PR #765, opened 2025-07-18
feat: API key authorization policy

hburn7:
This feature adds a new authorization policy which checks against the `X-Api-Key` header. This header & auth policy replace all instances of `[AllowAnonymous]` (with the exception of login/logout). This adds a protective layer against anonymous scraping.

So long as the correct header and secret is supplied, the request will succeed. otr-web is supplied this secret, but random users will not have access to this. Now, "anonymous" resources are accessible through platforms which we give the access to, rather than being completely open to the entire internet.


SourMongoose:
There seem to be a lot of line-length formatting changes - was anything changed in terms of expected code formatting?

hburn7:
Good eye - this would be my laptop rider settings getting in the way most likely.

hburn7:
CleanShot 2025-07-20 at 15 30 53@2x

Needed to disable this checkbox. Finally resolved.

